PR #2, opened 2019-10-05
added JSON cards, parser, and parser test

cvennel:


backend/src/game/cards/Card_JSON_Parser.py
```diff
@@ -0,0 +1,56 @@
+import json
+from enum import Enum
+from typing import List
+
+with open('backend\\src\\game\\cards\\KOTCards.json', "r") as json_file:
+    card_file = json.load(json_file)
+
+
+class CardType(Enum):
```
borisv13: Please extract this enum out into a separate class file.

backend/src/game/cards/Card_JSON_Parser.py
```diff
@@ -0,0 +1,56 @@
...
+    keep = 1
+    discard = 2
+    invalid = 99
+
+
+class Card(object):
```
borisv13: Please extract the Card object out as well. And make a separate file for the CardParser

backend/src/game/cards/Card_parsing_test.py
```diff
@@ -0,0 +1,124 @@
...
+            "card_type": "Keep",
+            "effect": "If you lose [health], roll a die for each [health] you lost. "
+                      "Each [heart] reduces the loss by 1[heart]",
+            "footnote": "(applicable to damages from both [attack] and Card effect)"
+        }
+    ]
+}
+
+
+def test_deck_multiple_cards():
```
borisv13: Tests should be in a test directory structure.
http://doc.pytest.org/en/latest/goodpractices.html

backend/src/game/cards/Card_JSON_Parser.py
cvennel: Extracted to CardType.py

backend/src/game/cards/Card_JSON_Parser.py
cvennel: Extracted to Card.py

backend/src/game/cards/Card_parsing_test.py
cvennel: moved to backend/test/game/cards_test/

backend/src/game/cards/Card_JSON_Parser.py
```diff
@@ -0,0 +1,56 @@
...
+        self.name = name
+        self.cost = cost
+        self.effect = effect
+        self.footnote = footnote
+        if str.lower(card_type) == "keep":
+            self.card_type = CardType.keep
+        elif str.lower(card_type) == "discard":
+            self.card_type = CardType.discard
+        else:
+            self.card_type = CardType.invalid
```
gpfaia: You must assert that card_type is type str otherwise you could potentially generate an error if someone passed in an int.  Also would suggest making this block it's own function

backend/src/game/cards/Card_JSON_Parser.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def card_details_str(self):
```
gpfaia: NIT: There has been a movement away from this type of concatenation for formatting
https://www.python.org/dev/peps/pep-3101/#string-methods

backend/src/game/cards/Card_JSON_Parser.py
```diff
@@ -0,0 +1,56 @@
...
+            assert isinstance(self.cost, int)
+            assert isinstance(self.card_type, CardType) and self.card_type != CardType.invalid
+        except AssertionError:
+            print("invalid card: " + self.card_details_str())
+            return False
+        return True
+
+
+def __deck_json_parser(json_deck):
+    deck: List[Card] = []
```
gpfaia: Invalid code

cvennel: invalid code? Are you referring to just line 43? I'm initializing an empty list called "deck" with a type hint of a List of Card objects. Could you elaborate on what I'm doing that's invalid?

backend/src/game/cards/Card_JSON_Parser.py
cvennel: Thanks! (does NIT stand for nitpick?)

backend/src/game/cards/Card_JSON_Parser.py
gpfaia: This was my fault, I was reading the code as if you were using py2.7 and type hints are a new feature as of 3.5.  The syntax threw me off

backend/src/game/cards/Card_JSON_Parser.py
cvennel: Ok I added more assertions to the is_valid method. Would that satisfy this requirement or should the assertions be in the init? Also extracted the block I think you referenced to it's own funciton

backend/src/game/cards/Card_JSON_Parser.py
gpfaia: Yep!

PR #7, opened 2019-10-09
Database model

rakhfa:
removed user_id and made username as primary key for Users table. Added basic tests for both models. @otaiwo70 @samridhprasad @borisv13 @kaubster @cvennel


backend/src/kot/models.py
```diff
@@ -1,14 +1,17 @@
...
+    email = models.CharField(primary_key=True, max_length=50, default="sample@drexel.edu",
+                             validators=[validate_email])
+    password = models.CharField(max_length=30, validators=[MinLengthValidator(8)])
     date_created = models.DateTimeField()
     date_modified = models.DateTimeField()
-    
+
+
 class Game(models.Model):
     GAME_RESULT = (
```
samridhprasad: This assignment looks ambiguous to me. Could you clarify what you mean by 'Y' / 'Yes' being the result of a game? 
@otaiwo70 @rakhfa

borisv13: I'd also like to see this be a recording of what place the player got, not just "did they win"

otaiwo70: refers to section below specifying whether a user is a winner or not a weather or undecided if the game is in progress

is_winner = models.CharField(max_length=1, choices=GAME_RESULT)

rakhfa: So far by design we are only tracking user's game statuses: win/lose/pending. If we want to track user's ranking in a game then we will have to change our db model accordingly. I will work on that.

PR #8, opened 2019-10-10
Player

borisv13:
Note that I did this Test Driven, that means each test was written before the code. The code was then written to meet the test case criteria. You can look at test_player.py and read the tests from top to bottom to see the order in which I built the player.

A couple notes
1) Player is getting too big. I'm tempted to refactor, but no convinced yet
2) I did not handle Victory Points = 20 win condition, and I don't believe that's the responsibility of the player to check, but rather a game engine type class
3) Based on that same logic, I think checking for player death might not belong inside the player either. We can refactor that easily later when we have such a game engine
4) No comments. Good code does not require comments unless you're doing some super sophisticated algorithm. Good naming is key.
5) I don't like that "constants" is lowercase but "Locations" is capitalized. However I checked PIP8 and the internet, and that seems to be correct? Someone let me know if that's wrong!


backend/src/game/locations.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class Locations(Enum):
+    OUTSIDE = 0
```
cvennel: Note that there are 3 locations in the game, but Tokyo and Tokyo Bay are for all intents and purposes the same. I don't think there are any situations where tracking a distinction between Tokyo and Tokyo Bay would be beneficial. While the rules discuss the two locations, it can be simplified to whenever there are more than 4 players, two monsters can be in Tokyo

cvennel: In other words, the absence of Tokyo Bay in the enum makes sense to me, but I don't know if this was the intent of the omission.

borisv13: It actually _does_ matter between Tokyo vs Tokyo Bay, because players move between the two.
I intentionally left that location out because we're not there yet. Let's get a 2 player game to work, first!

cvennel: Ah I re-read some rules and see where it matters now. Oops. Makes sense to focus on 2 player first though.

backend/src/game/player.py
```diff
@@ -14,6 +14,7 @@ def __init__(self):
         self.maximum_health = constants.DEFAULT_HEALTH
         self.current_health = self.maximum_health
         self.location = Locations.OUTSIDE
+        self.is_alive = bool(0)
```
cvennel: bool(0) is false isn't it? so init a player as not alive? seems backwards

borisv13: Derpity derp derp

backend/src/game/player.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def leave_tokyo(self):
+        self.location = Locations.OUTSIDE
+
+    def change_health(self, change_integer):
+        self.current_health += change_integer
+        if self.current_health > self.maximum_health:
+            self.current_health = self.maximum_health
+        if self.current_health <= constants.ZERO:
+            self.is_alive = bool(1)
```
gpfaia: In this statement, it should be bool(0).  Also for clarity and just good pratice, use the built-in False.

borisv13: image
What do?

backend/src/game/player.py
```diff
@@ -0,0 +1,29 @@
...
+        self.is_alive = bool(1)
+        self.victory_points = constants.ZERO
+
+    def move_to_tokyo(self):
+        self.location = Locations.TOKYO
+
+    def leave_tokyo(self):
+        self.location = Locations.OUTSIDE
+
+    def change_health(self, change_integer):
```
gpfaia: Due to the way the change health function and alive instance variable work, there is a known odd behavior here in that a player can be not alive but functions will still update the instance values.  Which means you could get things like a player could end up dying but if they were queued to do an action to add victory points, this would take place and the player could win the game.  There are side-effects here and there is definitely an argument that the game engine should handle all these circumstances not the player; however, it could make sense for a custom exception to be built that fires if an action is trying to be taken by a non-living player

backend/src/game/player.py
gpfaia: Right idea but your letting your test go against what would be code readability.  What you should do is make the value use the python built-ins but change the test to say.

` assert not player.is_alive`

The important thing I always remind people with PEP-8 is that it is a style "guide" not a directive. 

https://stackoverflow.com/a/14734103
https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html

backend/src/game/player.py
borisv13: Card abilities will eventually mess with a lot of this stuff, so I left the ability there. The game should be responsible for not declaring a dead player a winner :P

backend/src/game/player.py
borisv13: Coming from Java-lalaland. assert not is what I was looking for.
PEP8 wouldn't be a very good guide if there were situations where you HAD to go against it. I knew there'd be a solution.

backend/src/game/player.py
otaiwo70: For this section of code, wondering why
        self.maximum_health = constants.DEFAULT_HEALTH
        self.current_health = self.maximum_health
instead of simply  
        self.maximum_health = constants.DEFAULT_HEALTH
        self.current_health = constants.DEFAULT_HEALTH
the current solution although the same as my suggestion, was that done for readability or some other game logic

otaiwo70:
> Questions:
> 
> 1. Each player will pick a Monster to play. Should we add that in instance variable list? like name.
> 2. Player can also have list of special cards. Should we also have a list to keep track of that?
> 3. We can evaluate if a player is alive or dead based on current health (whether >0 or not). Do we need additional is_alive instance variable? Just thinking out loud. Not sure what is the best practice.

-------------------

1. We will definitely need to add the monster attached to player as an instance variable like "monster" instead
2. The cards would be better implemented as a class on their own, which we can init, shuffle, deal out and remove cards dealt out. The game will have an instance of the card class
3. Game class will have player instances 
#Above are thoughts going by the A has a B composition paradigm from Java, so a game has player so game will have instances of players, game has cards so an instance as well and so forth 

https://python-textbok.readthedocs.io/en/1.0/Object_Oriented_Programming.html

backend/src/game/player.py
gpfaia: This also can just be reduced further
`
self.maximum_health = self.current_health = constants.DEFAULT_HEALTH
`

borisv13: Thank you!

borisv13:
> Questions:
> 
> 1. Each player will pick a Monster to play. Should we add that in instance variable list? like name.
> 2. Player can also have list of special cards. Should we also have a list to keep track of that?
> 3. We can evaluate if a player is alive or dead based on current health (whether >0 or not). Do we need additional is_alive instance variable? Just thinking out loud. Not sure what is the best practice.

In the future, yes. This was more of a demonstration of what I expect in terms of testing and code quality.

borisv13:
Nothing, and I'm ok with that. Ideally, this class shouldn't even be
handling that logic. A game engine should be determining player death.

Vlek:
Player is getting too big. I'm tempted to refactor, but no convinced yet
    I did not handle Victory Points = 20 win condition, and I don't believe that's the responsibility of the player to check, but rather a game engine type class
    Based on that same logic, I think checking for player death might not belong inside the player either. We can refactor that easily later when we have such a game engine
    No comments. Good code does not require comments unless you're doing some super sophisticated algorithm. Good naming is key.
    I don't like that "constants" is lowercase but "Locations" is capitalized. However I checked PIP8 and the internet, and that seems to be correct? Someone let me know if that's wrong!


> 1. Player is getting too big. I'm tempted to refactor, but no convinced yet

This is a stub of the full functionality of the player. While there's likely to be a lot here, I still think that it logically makes sense to be contained in the player class.

> 2. I did not handle Victory Points = 20 win condition, and I don't believe that's the responsibility of the player to check, but rather a game engine type class

I agree, this is something the game should deal with. We can however make it easier for the game by having the method that changes victory points return the number of VP that the user has.

> 3. Based on that same logic, I think checking for player death might not belong inside the player either. We can refactor that easily later when we have such a game engine

I don't necessarily agree with this. The player health change should be able to take a positive or negative number. If the player gets down to zero, that's player death. Then we could have a method like `isDead` to be able to see whether that is the case to know whether to allow healing again.

The problem I foresee with this is that there's a card that, if active, a player can lose all of their cards, energy, and victory points in order to come back at full health, essentially as though they've just started the game. However, for this, we could just re-instantiate the player object anew instead of attempting to reset all of the state to get them back to square one.

5. I don't like that "constants" is lowercase but "Locations" is capitalized. However I checked PIP8 and the internet, and that seems to be correct? Someone let me know if that's wrong!

I believe this is how you would use constants because the constant itself would be all upper case but modules have a separate naming convention as you stated.


Other comments:
- Healing does not consider whether the player is dead or is in Tokyo/Tokyo Bay, which is not allowed.
- I don't believe that `constants.ZERO` is better than simply `0`, unless there's something I'm missing regarding the actual value of zero ever changing within the context of the game.

borisv13:
Single responsibility principle.
It's not the player's responsibility to declare itself dead. Only the game
can know and decide that based on the cards they have (for example revival
or something else tricky).
Healing doesn't take anything into account for the same reason - it's not
for the player the keep track of the rules of the game, but rather the game
to not heal a player that is inside Tokyo.

PR #10, opened 2019-10-12
Adding dice.py!

Vlek:
Ran some tests using pytest showing that both rolling and multiple
rolls work correctly.

### DIFF
--- a/backend/src/game/dice.py
+++ b/backend/src/game/dice.py
@@ -1,28 +1,19 @@
-"""
-This is a dice module for the game King of Tokyo
-
-The dice can have the following face values:
-    - 1 - 3: These are the amount of victory points earned
-    - attack: Depending on the player's location, hurts either monsters
-    in or outside Tokyo
-    - heal: Heals the player up to a maximum of 10 points
-    - energy: Gives the player an energy cube to spend on upgrade cards
-"""
-
 import random
+from enum import Enum
 
-DICE_VALUES = ["1", "2", "3", "attack", "heal", "energy"]
+
+class DieValue(Enum):
+    ONE = 0
+    TWO = 1
+    THREE = 2
+    ATTACK = 3
+    HEAL = 4
+    ENERGY = 5
 
 
 def roll() -> str:
-    """
-    Returns back a randomly selected dice value.
-    """
-    return random.choice(DICE_VALUES)
+    return random.choice(list(DieValue))
 
 
-def rolls(num: int) -> [str]:
-    """
-    Returns back a number "num" of randomly selected dice values in a list
-    """
+def roll_many(num: int) -> [str]:
     return [roll() for _ in range(num)]
--- a/backend/src/game/dice.py
+++ b/backend/src/game/dice.py
@@ -1,0 +1,28 @@
+"""
+This is a dice module for the game King of Tokyo
+
+The dice can have the following face values:
+    - 1 - 3: These are the amount of victory points earned
+    - attack: Depending on the player's location, hurts either monsters
+    in or outside Tokyo
+    - heal: Heals the player up to a maximum of 10 points
+    - energy: Gives the player an energy cube to spend on upgrade cards
+"""
+
+import random
+
+DICE_VALUES = ["1", "2", "3", "attack", "heal", "energy"]
+
+
+def roll() -> str:
+    """
+    Returns back a randomly selected dice value.
+    """
+    return random.choice(DICE_VALUES)
+
+
+def rolls(num: int) -> [str]:
+    """
+    Returns back a number "num" of randomly selected dice values in a list
+    """
+    return [roll() for _ in range(num)]
--- a/backend/test/game/test_dice.py
+++ b/backend/test/game/test_dice.py
@@ -1,18 +1,7 @@
-# import pytest
+import pytest
+
 import backend.src.game.dice as dice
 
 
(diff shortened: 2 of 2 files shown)


backend/src/game/dice.py
```diff
@@ -0,0 +1,28 @@
...
+
+import random
+
+DICE_VALUES = ["1", "2", "3", "attack", "heal", "energy"]
+
+
+def roll() -> str:
+    """
+    Returns back a randomly selected dice value.
+    """
```
cvennel: In the sample pull request from @borisv13 he states, "4. No comments. Good code does not require comments unless you're doing some super sophisticated algorithm. Good naming is key." I was under the impression that means as a rule we would try to avoid comments like this

borisv13:
Correct. Sorry to be anal but I'm here to teach and one of those things is
Clean Code.
Clean code is self documenting.

backend/src/game/dice.py
```diff
@@ -0,0 +1,28 @@
+"""
```
gpfaia: Some of this information may not be bad for a docstring; however, I believe the dice should be an established class not just a module.

backend/src/game/dice.py
```diff
@@ -0,0 +1,28 @@
...
+    - 1 - 3: These are the amount of victory points earned
+    - attack: Depending on the player's location, hurts either monsters
+    in or outside Tokyo
+    - heal: Heals the player up to a maximum of 10 points
+    - energy: Gives the player an energy cube to spend on upgrade cards
+"""
+
+import random
+
+DICE_VALUES = ["1", "2", "3", "attack", "heal", "energy"]
```
gpfaia: I understand why you set everything to a string type, that way it is consistent; however, thinking about how the game will have to evaluate the dice value (eg. if there are 4 "2"s, you will have to cast that string into an int to understand what value it is but you only need to do that for integers, your function would not need to do this for attack, heal, and energy.  Since this is the case it makes sense for the integer values to remain as ints that may it is an easier thing to keep track of later on.  It may make sense to go even a step further and declare types for each of these dices

borisv13:
Enums get assigned values, so that solves that problem as well, right?

backend/src/game/dice.py
Vlek: When you do that though, then you have to use it as follows:
```d = Dice()
d.roll()
```

It's also not helpful as a class because we're not having to save state or anything. I had originally wrote it as a class and it just didn't feel right when I went to use it.

backend/src/game/dice.py
Vlek: I vehemently disagree regarding the commenting, especially with the code sample that was provided in Player.py.

What was given looked much more like Java code, especially with the use of enums and constants, versus something that would be more obvious on its face.

Comments should always be included, especially docstrings, which is a Pythonic standard.

borisv13:
I feel like there should be an enum called Die with the values on any given die. And then this class should be more of a DieRoller.

borisv13:
I just want readable code. If docstrings are standard, then keep them in.
More than happy to be overruled.

borisv13:
I didn't think that's a Java thing. Enums exist in python for a reason, too.
When the game wants to check if a certain die is a heart, we shouldn't have
to compare to "heart". That's what enums exist for. Fixed values that are
easily readable.

borisv13:
http://xion.io/post/code/python-enums-are-ok.html

Shrug
Call me out if I'm wrong! I'm always more than happy to concede if I'm
missing something.

backend/src/game/dice.py
Vlek: If I used an enum however as the professor is suggesting, then I am going to have to do some type coercion still:

When we get to doing actual game coding, it's going to be just as easy to do something like:
``` Python
current_player.heal(sum([1 if die == "heart" else 0 for die in rolls]))
```
As it would to do:
``` Python

class FaceValues(enum):
    One = 0
    Two = 1
    Three = 2
    Attack = 3
    Heart = 4
    Energy = 5

current_player.get_victory_points(sum([int(die) + 1 if die in [FaceValues.One, FaceValues.Two, FaceValues.Three] else 0 for die in rolls]))
```

borisv13:
Readability + duplication.
General rule of thumb: don't make a dev write the same literal twice.
"heart" is an example of that. That should only be written in literal form
once in the MT and once in the UI. The rest should be calls to a
constant/enum.

borisv13:
From the Google python style guide:

A function must have a docstring, unless it meets all of the following
criteria:

   - not externally visible
   - very short
   - obvious

I stand corrected.

backend/src/game/dice.py
borisv13: Apparently when I reply by email, the reply doesn't go to the conversation. It just gets dumped at the bottom of the review lol. Check out below

borisv13: I don't like Face Values as I don't think a new dev would think to look for that name not something like Die or DieFaceValues

backend/src/game/dice.py
cvennel: copying reply from @borisv13  into conversation to resolve conversation:

> From the Google python style guide: A function must have a docstring, unless it meets all of the following criteria: - not externally visible - very short - obvious I stand corrected.

backend/src/game/dice.py
cvennel: Personally, I like the enum version.

PR #11, opened 2019-10-13
Early game engine

gpfaia:
Some extremely early logic to try and map out some of the game state.

- Take turns is a stub function as we have not merged in the dice roll.

- It does not utilize our player class yet as was unsure of how we wanted to manage that so it is just inserting in a generic "player"

- The yield may need more explicitness because need to be careful when changing an active list

### DIFF
--- a/backend/src/game/engine.py
+++ b/backend/src/game/engine.py
@@ -7,18 +7,19 @@ class GameEngine():
         self.players = []
         self.alive_players = []
         self.dead_players = []
+        self.game_status = Status.GAME_OVER
 
     def add_player(self, player):
         self.players.append(player)
 
     def begin_game(self):
         self.alive_players = self.players
+        self.game_status = Status.ACTIVE
 
     def play_game(self):
-        game_status = Status.ACTIVE
-        for player in self.yield_next_player(game_status):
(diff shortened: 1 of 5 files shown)


backend/src/game/engine.py
```diff
@@ -0,0 +1,43 @@
...
+                self.validate_if_winner(player)
+                if not self.game_status:
+                    break
+                self.set_dead_players()
+
+    def take_turn(self, player):
+        pass
+
+    def validate_if_winner(self, player):
+        if player.victory_points == 20:
```
borisv13: Please add 20 to constants.py

backend/src/game/engine.py
```diff
@@ -0,0 +1,43 @@
...
+            self.game_status = Status.GAME_OVER
+
+    def set_dead_players(self):
```
borisv13: How do you see this method working? When is this triggered, how is it used?
The way I had it in my head:
At the end of every turn, the dice are handled. Damage is dealt. At this point, players would be marked dead if conditions are met. In which case I don't think we need this method. All players start as alive and then when they're marked dead, they're moved.
But this doesn't mean that's the way it has to be. Looking forward to what's in your head.

PR #12, opened 2019-10-14
Deck service

cvennel:
Created a Deck class and DeckHandler class. Together these manage interactions with the draw and discard piles and the store. 

some notes:
-buying a card for now just returns a card object. A future iteration will include making sure a player can actually afford it and deduct energy etc. for now this let me keep my portion completely separate since player hasn't implemented energy yet. 

-the discard pile automatically gets shuffled and becomes the draw pile when trying to look at the top card of an empty deck

-trying to draw a card from an empty draw pile with an no discard pile to shuffle just raises an exception for now


backend/src/game/cards/card.py
```diff
@@ -0,0 +1,35 @@
...
+        self.effect = effect
+        self.footnote = footnote
+        self.card_type = set_card_type(card_type)
+
+    def card_details_str(self):
+        return "\nname: {0}\ncost: {1}\ntype: {2}\neffect: {3}\nfootnote: {4}".format(self.name, self.cost,
+                                                                                      self.card_type,
+                                                                                      self.effect, self.footnote)
+
+    def is_valid(self):
```
borisv13: Do we need this? Isn't validation taken care of in the parser?

backend/src/game/cards/card_type.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+
+class CardType(Enum):
```
borisv13: All caps for the enum words, KEEP, etc

backend/src/game/cards/deck_handler.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @property
+    def store(self):
+        return self.__card_store
+
+    @property
+    def discard_pile(self):
+        return self.__discard_pile
+
+    def get_top_draw_pile_card(self):
```
borisv13: Tests for each of the possible branches in this method, please

backend/src/game/cards/card.py
cvennel: The parser uses this to check if the card is valid before appending the deck it's creating.

backend/src/game/cards/card_type.py
cvennel: Then we should update the locations enum 
`
from enum import Enum


class Locations(Enum):
    OUTSIDE = 0
    TOKYO = 1

`

cvennel: @borisv13  correct me if I'm wrong, but does this mean in Python enums should be named like classes? https://www.python.org/dev/peps/pep-0435/

backend/src/game/cards/deck_handler.py
cvennel: The tests are already in deck_handler_test.py

normal ->test_buy_single_card
if draw pile is empty -> test_auto_reshuffle
if discard is also empty-> test_run_out_of_cards

backend/src/game/cards/card_type.py
borisv13: https://stackoverflow.com/questions/21359130/should-3-4-enums-use-upper-case-with-underscores
https://docs.python.org/3/library/enum.html#creating-an-enum

I thought I did do Locations as all caps... am I crazy?

cvennel: OH! I misunderstood. I thought you meant LOCATION.TOKYO not Location.TOKYO.  Now I get it. I'll fix that

cvennel: Fixed it

kaubster:
Looks good,

Is it possible to add the number of cards to draw to test_draw_card?

for example:

def test_draw_card(deck,numToDraw)

backend/src/game/constants.py
```diff
@@ -1,2 +1,13 @@
+import os
+
+# directories
+CARDS_DIRECTORY = os.path.dirname(os.path.abspath(__file__))
+
+# ints
 DEFAULT_HEALTH = 10
 ZERO = 0
+CARD_STORE_SIZE = 3
+JSON_CARD_COUNT = 66
```
borisv13: Do we need this in prod code? 
It looks like this is only used in the tests, in which case this should exist in a constants file that lives with or even in the test file.

cvennel: Oh duh. I'll fix that this week

cvennel:
@borisv13 yes, card store size is prod. I fixed it quickly but was having a weird time with GIT not doing what I wanted. Should be good now.

cvennel:
@kaubster I don't know what the benefit of that test would be. All that class is responsible for is drawing a single card. After that, the deck handler is managing drawing cards multiple times and handling what to do when running out of cards

backend/test/game/cards_test/deck_handler_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+@pytest.fixture(autouse=True)
+def deck_handler():
+    deck_handler: DeckHandler = DeckHandler()
+    return deck_handler
+
+
+def test_deck_handler_init(deck_handler):
+    print(len(deck_handler.store))
+    print(len(deck_handler.draw_pile))
```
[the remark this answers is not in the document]
cvennel: I'll use those while I'm developing the test to confirm that what I think is happening is what's actually happening. When the test fails it gives me some more information. Helped me learn in Python the difference between list1 = list2 and list1 =list2.copy() which was a pain. Normally I remove the print statements at the end when I clean up the code. In this case I missed them.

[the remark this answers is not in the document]
cvennel: Removed

PR #14, opened 2019-10-17
Card hierarchy

borisv13:
Chastise my late night coding, please!
Also, I like keeping the test file WITH the code file, rather than replicating the /test folder structure to be the same as the /src folder structure. How do we feel about that? Stackoverflow supported/enabled my decision lol


borisv13:
> Good place to start but I'll also attend other methods to handle other stages of effects..
> 
> For example pre-roll effects pre-purchase effects pre-attack effects and so on?
> 
> Basically look at the effects and come up with an interface that all cards implement.

TDD - I don't add anything I don't have a test case for yet.
Just like I didn't add KeepCard.
It's implied these things will come to exist, but come on they take seconds to add - so add them when the time comes! :)

backend/src/game/newcards/discard_cards/victory_point_manipulation_cards/test_apartment_building.py
```diff
@@ -0,0 +1,29 @@
...
+    assert player.victory_points == 3
+
+
+def test_apartment_building_costs_5_energy():
+    assert ApartmentBuilding().cost == 5
+
+
+def test_apartment_building_has_effect():
+    effect = ApartmentBuilding().effect
+    assert effect and not effect.isspace()
```
borisv13: @gpfaia Is there a better way to write this?

gpfaia:
Is this a replacement for the other cards folder that already exists because some of the code here is repetitive?

backend/src/game/newcards/new_card.py
```diff
@@ -0,0 +1,9 @@
+from abc import ABC
+
+
+class NewCard(ABC):
```
gpfaia: I am not in love with this class name here but I do not currently have an alternative.  Maybe just GameplayCard or something more to that effect

backend/src/game/newcards/new_card.py
```diff
@@ -0,0 +1,9 @@
...
+    def __init__(self, name, cost, effect, footnote):
```
gpfaia: The value of footnote should be `footnote=None`, this is due to the fact that often there will not always be a footnote so it can be an optional field. If the argument is but it will make it easier to display if it's an empty string instead of a None, this is bad design as you are using information relevant to visualizations about how you create the underlying data structures which means you're solving a problem before it exists and only works if the visualization does not change.

backend/src/game/newcards/discard_cards/victory_point_manipulation_cards/apartment_building.py
```diff
@@ -0,0 +1,10 @@
+from backend.src.game.newcards.discard_card import DiscardCard
+from backend.src.game.newcards.new_card import NewCard
+
+
+class ApartmentBuilding(DiscardCard):
+    def __init__(self):
+        NewCard.__init__(self, "Apartment Building", 5, "+ 3[Star]", "")
```
gpfaia: This line and all others without footnotes could then drop the final argument in the init as it would be no longer necessary and would lead to cleaner code overall

backend/src/game/player.py
```diff
@@ -19,10 +19,10 @@ def change_health(self, change_integer):
         self.current_health += change_integer
         if self.current_health > self.maximum_health:
             self.current_health = self.maximum_health
-        if self.current_health <= constants.ZERO:
+        if self.current_health <= constants.DEATH_HIT_POINT:
             self.is_alive = False
 
     def change_victory_points(self, change_integer):
```
gpfaia: This is outside the scope of this PR but these functions "change_*" would imply that you are taking the value and changing it to the value that is given.  eg. If you victory_points = 10 and you see change_victory_points(3), my mind would think it is changing the value to 3 not adding to it to make it 13.  Would need to change up the language to make this much clearer

backend/src/game/newcards/discard_cards/victory_point_manipulation_cards/test_apartment_building.py
gpfaia: This is similar to my comment about footnote.  If it is determined that there are elements that will not have effects, then you should use the None in the __init__.  If it will always have a value then it makes sense to write just this since that is what you want to test, you are not writing a generic test for this one.

`assert effect == "+ 3[Star]"`

In the future if writing a more generic
https://stackoverflow.com/a/45968236

kaubster:
I like the abstract approach twards representing cards. Ok with this PR.

Hesitating to close given the remaining comment by @gpfaia.

Future: At somepoint integration tests between loading the cards and generating these abstract instances might be beneficial (demonstrate integration).

backend/src/game/newcards/discard_cards/victory_point_manipulation_cards/test_apartment_building.py
borisv13: I don't want to assert on the string value of the effect. If we go down that path, it'll be painful. Pull Requests are there to make sure the strings for the cards are correct. Duplicating that in the test doesn't add value.
Good to know about the =None, I will do that.

backend/src/game/newcards/new_card.py
borisv13: It will be Card, it's temporarily NewCard while we haven't retired Chris' parser code yet.

backend/src/game/player.py
borisv13: Ok I can see that, no one brought it up in the Player PR.
I'll see what I can do, but if you have a better name suggestion, please write.

gpfaia: The best I can think of off the top of my head would just be 'change_value_by' or 'update_value_by', would keep the same functionality under the covers but the implication is your changing it BY that value not TO it

backend/src/game/newcards/discard_cards/victory_point_manipulation_cards/test_apartment_building.py
gpfaia: If that is the case, you can probably get rid of the footnote test case then as it won't be needed.

For this test,  I think it would be safe to  just adding an assert type str for effect into the NewCard class's init method.  The secondary condition is not necessary as you're basically repeating code.

I think what you may have been looking for is
`assert effect and isinstance(effect, str)`

borisv13: > Is this a replacement for the other cards folder that already exists because some of the code here is repetitive?

Yes

backend/src/game/player.py
samridhprasad: I like the `update_value_by` approach, makes it clear to handle increments/decrements

PR #16, opened 2019-10-18
 added dice_handler and tests

cvennel:
Created logic to roll and re-roll dice. Re-rolling is done by passing a list of indexes or a single int describing which dice to re-roll (0-5 for the original 6 dice)


cvennel:
Corrected unnecessary print statements

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,60 @@
...
+import backend.src.game.dice as dice
+
+
+@pytest.fixture(autouse=True)
+def dice_handler():
+    dice_handler = DiceHandler()
+    return dice_handler
+
+
+def test_start_turn(dice_handler):
```
borisv13: assert that there's 6 die

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,60 @@
...
+    for die in dice_handler.dice_values:
+        assert die in dice.DieValue
+
+
+def test_re_roll_one_die(dice_handler):
+    dice_handler.start_turn()
+    start_value = dice_handler.dice_values[0]
+    roll_changed_value = False
+    # since technically possible re-roll returns same roll, run 10 times and quit on successful result
+    for _ in range(10):
```
borisv13: Please don't rely on random within tests.
Mock out the roll() to force the die to roll to a specific value you can assert it has changed to.
1. Mock out rollMany() to return a set, like 1,2,3,atk,heal,energy
2. Mock out roll() to return something fixed, like energy
3. Call re_roll_dice(0) 
4. assert energy,2,3,atk,heal,energy

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,60 @@
...
+            break
+    assert roll_changed_value
+
+
+def test_re_roll_all_dice(dice_handler):
+    dice_handler.start_turn()
+    start_values = dice_handler.dice_values.copy()
+    dice_handler.re_roll_dice([0, 1, 2, 3, 4, 5])
+    all_elements_are_same = True
+    for i in range(0, len(dice_handler.dice_values)-1):
```
borisv13: 1. Mock out 1 set
2. Mock out 2 set
3. Get set 1
4. Call for reroll
5. Assert 2nd set is set 2 you set up

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,60 @@
...
+    assert not all_elements_are_same
+
+
+def test_add_die(dice_handler):
+    dice_handler.start_turn()
+
+    # assert there's another die
+    start_die_count = len(dice_handler.dice_values)
+    dice_handler.add_bonus_die()
+    assert len(dice_handler.dice_values) > start_die_count
```
borisv13: I think you can assert that the len is exactly start_die_count + 1, which is better than >

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # assert the last die can be rerolled
+    last_die_index = len(dice_handler.dice_values) - 1
+    bonus_die_initial_val = dice_handler.dice_values[last_die_index]
+    bonus_die_changed = False
+    # since technically possible re-roll returns same roll, run 10 times and quit on successful result
+    for i in range(10):
```
borisv13: Split into its own test
use mocks.

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,24 @@
+import backend.src.game.dice as dice
+import backend.src.game.constants as constants
+
+
+class DiceHandler(object):
```
Vlek: There's no reason in Python3 to inherit from object, see: https://stackoverflow.com/questions/4015417/python-class-inherits-object

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,24 @@
...
+    dice_values = list()
```
Vlek: Use a list literal instead of this. It's the idiomatic way and it's also faster: https://softwareengineering.stackexchange.com/questions/319547/literals-versus-instantiating-by-name-lists-and-dicts-in-python

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,24 @@
...
+    __re_rolls_left = constants.DEFAULT_RE_ROLL_COUNT
```
Vlek: There's no reason here to use underscores. You're not creating a object-level variable that couldn't possibly be touched from the outside. The double underscore (which is different from dunder) also implies that it's a class level variable that could change at any given moment so it shouldn't be used outside of the internals for this object.

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def start_turn(self):
+        self.dice_values = dice.roll_many(constants.DEFAULT_DICE_COUNT)
+        self.__re_rolls_left = constants.DEFAULT_RE_ROLL_COUNT
+
+    def re_roll_dice(self, indexes_of_dice_to_re_roll):
+        if self.__re_rolls_left > 0:
```
Vlek: If you're not going to raise an exception for this and go on, instead of nesting everything, you could just return out at this point.

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,24 @@
...
+            if type(indexes_of_dice_to_re_roll) == int:
+                self.dice_values[indexes_of_dice_to_re_roll] = dice.roll()
+
+            elif type(indexes_of_dice_to_re_roll) == list:
+                for val in indexes_of_dice_to_re_roll:
+                    self.dice_values[val] = dice.roll()
```
Vlek: For this, I would strongly suggest against allowing different types for the parameter. Instead, I would suggest only allowing a list of type integer. If there's only one, then they'll give a list with only one integer.

You may also want to consider making this two separate methods, one for doing a single roll with an integer and this method which allows a list.

Vlek: There's also no check here to make sure that what the client is sending makes sense. If they send an integer that's outside of the range of the dice, then it's going to throw an out of index error and explode.

backend/src/game/constants.py
```diff
@@ -7,6 +7,8 @@
 DEFAULT_HEALTH = 10
 ZERO = 0
 CARD_STORE_SIZE_LIMITER = 3
+DEFAULT_RE_ROLL_COUNT = 2
+DEFAULT_DICE_COUNT = 6
```
Vlek: I was already overruled by the professor on this, but I want to make a note that I believe that all things dice should reside in the dice file. These I feel should be class-level variables.

https://www.digitalocean.com/community/tutorials/understanding-class-and-instance-variables-in-python-3

backend/src/game/dice_handler.py
borisv13: 100% so much yes

backend/src/game/dice_handler.py
borisv13: I don't think we need 2 methods. Just accept a list of ints and move on.
We do need error handling somewhere along the lines.

backend/src/game/constants.py
borisv13: Appreciate the relentlessness, but still overruled. These should not be owned by the Dice class or the DiceHandler class. They shouldn't even be used inside of the Dice or DiceHandler class (which will be refactored later).
The engine responsible for using DiceHandler should pass these values in as input params for a given player.
something like:
dice_handler(DEFAULT_RE_ROLL_COUNT, DEFAULT_DICE_COUNT)
or if the player has special cards, maybe dice_handler(DEFAULT_RE_ROLL_COUNT+1, DEFAULT_DICE_COUNT)
The responsibility of knowing the rulebook should not be in Dice or DiceHandler is the point.

borisv13: @cvennel Actually we can do this refactor now, save us effort later. Please add these two constants as input params. I *think* dice_count can just come in to start_turn(), but reroll_count should be class level? Not sure though

backend/test/game/dice_handler_test.py
cvennel: oh duh

backend/test/game/dice_handler_test.py
cvennel: Thanks for the steps!

backend/test/game/dice_handler_test.py
cvennel: agreed

backend/src/game/dice_handler.py
cvennel: Thanks for including the references to read, it's been really helpful

backend/src/game/dice_handler.py
cvennel: Again, thanks for including the references to read, it's been really helpful!

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,26 @@
+import backend.src.game.dice as dice
+import backend.src.game.constants as constants
+
+
+class DiceHandler:
+    dice_values = []
```
gpfaia: Class variables although are able to be defined as non-static, should typically be static.  Also this would be getting shared across instances and I am unsure how you are planning on using the dice handler but I do not think that is the best way of storing this

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,26 @@
...
+    re_rolls_left = constants.DEFAULT_RE_ROLL_COUNT
+
+    def start_turn(self):
+        self.dice_values = dice.roll_many(constants.DEFAULT_DICE_COUNT)
+        self.re_rolls_left = constants.DEFAULT_RE_ROLL_COUNT
+
+    def re_roll_dice(self, indexes_of_dice_to_re_roll):
```
gpfaia: This should really only accept one type of input and then you would not need type checking, which exposes you to more possible issuses later on.

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,26 @@
...
+        if self.re_rolls_left <= 0:
+            return
+
+        if type(indexes_of_dice_to_re_roll) == int:
+            self.dice_values[indexes_of_dice_to_re_roll] = dice.roll()
```
gpfaia: How is there a minus re_reoll in line 23 but not here, I think this function could be refactored

backend/src/game/dice_handler.py
cvennel: Thanks!

backend/src/game/dice_handler.py
cvennel: Ah you're right. That was a big miss. This was already refactored anyways.

backend/src/game/dice_handler.py
cvennel: I'm trying to follow up on this one to make sure I understand but having trouble finding exactly what you're referencing.  Generally If I'm understanding correctly, I'm finding static variables are not really a thing in python like other languages. Do you mean the difference between class and instance variables? I found this, https://syntaxdb.com/ref/python/class-variables and realized I didn't understand the difference between putting things in the __init__ method and where I put it.

cvennel: I moved the values to be instance values instead of class variables. I think that addresses your concerns

backend/src/game/dice_handler.py
```diff
@@ -0,0 +1,29 @@
...
+    def re_roll_dice(self, indexes_of_dice_to_re_roll):
+        if self.re_rolls_left <= 0:
+            return
+
+        try:
+            for val in indexes_of_dice_to_re_roll:
+                self.dice_values[val] = dice.roll()
+            self.re_rolls_left -= 1
+
+        except IndexError:
```
gpfaia: This index error only needs to wrap line 19, it does not have to be this far away

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,67 @@
...
+import backend.src.game.dice.dice as dice
+from mock import Mock
+
+six_mock_dice_values_a = [dice.DieValue.ONE, dice.DieValue.TWO, dice.DieValue.THREE,
+                          dice.DieValue.ATTACK, dice.DieValue.ENERGY, dice.DieValue.HEAL]
+
+six_mock_dice_values_b = [dice.DieValue.ATTACK, dice.DieValue.ENERGY, dice.DieValue.HEAL,
+                          dice.DieValue.ONE, dice.DieValue.TWO, dice.DieValue.THREE]
+
+dice.roll = Mock(return_value=dice.DieValue.ATTACK)
```
borisv13: If you can, set the Mock Attack dice to a more appropriate name, such as MOCK_DIE (constant)
Then in your assertions, you can assert on the value being MOCK_DIE

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+from backend.src.game.dice_handler import DiceHandler
+import backend.src.game.dice.dice as dice
+from mock import Mock
+
+six_mock_dice_values_a = [dice.DieValue.ONE, dice.DieValue.TWO, dice.DieValue.THREE,
+                          dice.DieValue.ATTACK, dice.DieValue.ENERGY, dice.DieValue.HEAL]
+
+six_mock_dice_values_b = [dice.DieValue.ATTACK, dice.DieValue.ENERGY, dice.DieValue.HEAL,
```
borisv13: I don't see this being used anywhere.

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,67 @@
...
+    with pytest.raises(TypeError):
+        assert dice_handler.re_roll_dice(["BAD VALUE"])
+
+
+def test_add_die(dice_handler):
+    start_die_count = len(dice_handler.dice_values)
+    dice_handler.add_bonus_die()
+    assert len(dice_handler.dice_values) == start_die_count + 1
+    last_die_index = len(dice_handler.dice_values) - 1
+    assert dice_handler.dice_values[last_die_index] == dice.DieValue.ATTACK
```
borisv13: Rather than asserting = dice.DieValue.ATTACK, which may make someone reading the test ask "why attack?" it would be more ideal to assert = MOCK_DIE, such that it's obvious that the type of die is irrelevant, it's just that it's the die that was expected.

backend/test/game/dice_handler_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_roll_bonus_die(dice_handler):
+    dice_handler.add_bonus_die()
+    last_die_index = len(dice_handler.dice_values) - 1
+    bonus_die_initial_val = dice_handler.dice_values[last_die_index]
+    assert bonus_die_initial_val == dice.DieValue.ATTACK
```
borisv13: Same here

backend/test/game/dice_handler_test.py
cvennel: I like that idea. implemented

backend/test/game/dice_handler_test.py
cvennel: oops. fixed

backend/src/game/dice_handler.py
cvennel: I think it needs to wrap line 20 as well. I don't want to cost a turn if we don't actually get a valid result. Was this comment more about the extra line space I have on line 21? Cause that seems wrong

gpfaia:
I believe this will need a rework because of the python reorganization

backend/game/dice_handler.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self):
+        self.dice_values = []
+        self.re_rolls_left = 0
+
+    def start_turn(self, starting_dice_count, re_roll_count):
+        self.dice_values = dice.roll_many(starting_dice_count)
+        self.re_rolls_left = re_roll_count
+
+    def re_roll_dice(self, indexes_of_dice_to_re_roll):
+        if self.re_rolls_left <= 0:
```
gpfaia: I can see this if statement being it's own function as later on things may change the amount of rerolls left.  If not wanting to make it it's own function that returns a boolean, you can just change the logic to
`if self.re_rolls_left > 0`
that way you do not have to have an awkward return in the middle which kind of muddles the purpose of the function

cvennel: I like the idea of flipping the logic for legibility.

cvennel: applied change

backend/game/dice_handler.py
```diff
@@ -0,0 +1,25 @@
+import game.dice.dice as dice
+
+
+class DiceHandler:
+    def __init__(self):
+        self.dice_values = []
+        self.re_rolls_left = 0
+
+    def start_turn(self, starting_dice_count, re_roll_count):
```
Vlek: I don't think the dice handler should have anything regarding turns? This notion of a turn should probably be in what calls DiceHandler.

backend/game/dice_handler.py
```diff
@@ -0,0 +1,25 @@
...
+        self.dice_values = dice.roll_many(starting_dice_count)
+        self.re_rolls_left = re_roll_count
+
+    def re_roll_dice(self, indexes_of_dice_to_re_roll):
+        if self.re_rolls_left > 0:
```
Vlek: An earlier point that was made was that we can return out here if it's the case that it's zero or less.

backend/game/dice_handler.py
```diff
@@ -0,0 +1,25 @@
...
+            try:
+                for val in indexes_of_dice_to_re_roll:
+                    self.dice_values[val] = dice.roll()
+                self.re_rolls_left -= 1
```
Vlek: With it like this, you could potentially get SOME of the dice rerolled and then have the last one be an index error or type error to not have the reroll counter decrement.

backend/game/dice_handler.py
```diff
@@ -0,0 +1,25 @@
...
+            except IndexError:
+                raise IndexError("Re-roll index outside of dice_values range")
+            except TypeError:
+                raise TypeError("Re-roll index must be passed as list of ints")
+
+    def add_bonus_die(self):
```
Vlek: This should probably take an integer for the number of bonus die otherwise we would have to call this a number of times.

backend/game/dice_handler.py
borisv13: first_roll or initial_roll probably better name

cvennel: good point

backend/game/dice_handler.py
cvennel: this used to be 
```python
if self.re_rolls_left < 0
    return
```

but I thought Gerardo said the return muddled the purpose of the method. Maybe I'm misunderstanding.

cvennel: switched it back

backend/game/dice_handler.py
cvennel: Ooh so basically pass reroll [ 3,4,5,19] and you get free re-rolls. Good point.

backend/game/dice_handler.py
cvennel: done

backend/game/dice_handler.py
cvennel: updated method to ignore bad indexes

PR #17, opened 2019-10-18
Created socket.io server for Lobby

CameronBaffuto:
Created the server for players to connect to lobby, still need to add interface

### DIFF
--- a/frontend/src/containers/Lobby.js
+++ b/frontend/src/containers/Lobby.js
@@ -1,18 +1,21 @@
 import React from 'react';
 import "./Lobby.css";
+import openSocket from "socket.io-client";
+
+const socket = openSocket("http://localhost:3000");
 
 const LobbyLayout = (props) => {
     return (
         <div className="container">
             <div className="row">
                 <div className="col-sm">
-                    
+
                 </div>
                 <div className="col-sm">
                     Your In the Lobby...
                 </div>
                 <div className="col-sm">
-                    
+
                 </div>
             </div>
         </div>
@@ -20,5 +23,3 @@ const LobbyLayout = (props) => {
 }
 
 export default LobbyLayout;
-
-
--- a/frontend/src/server/Room.js
+++ b/frontend/src/server/Room.js
@@ -1,0 +1,33 @@
+import React from 'react';
+
+var app = require("express")();
+var http = require("http").Server(app);
+var io = require("socket.io")(http);
+
+const port = 3000;
+
+app.get("/", function(req, res) {
+  res.sendFile(__dirname + "/index.html");
+});
+
+io.on("connection", function(socket) {
+  console.log("a user connected");
+
+  socket.on("join_room", room => {
+    socket.join(room);
+  });
+
+    http.listen(port, function() {
+  console.log(`listening on *:${port}`);
+});
+
+socket.emit("new_visitor", visitor);
+
+socket.on("visitors", visitors => {
+  this.setState({
+    visitors: visitors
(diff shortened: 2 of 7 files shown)


frontend/src/containers/Lobby.js
```diff
@@ -1,24 +1,25 @@
 import React from 'react';
 import "./Lobby.css";
+import openSocket from "socket.io-client";
+
+const socket = openSocket("http://localhost:3000");
```
Vlek: Instead of hard coding this, you may want to add this to a config somewhere. Eventually, when we start getting things hooked up, you don't want to have to keep changing this here to go back and forth.

frontend/src/containers/Lobby.js
```diff
@@ -1,24 +1,25 @@
...
 const LobbyLayout = (props) => {
     return (
         <div className="container">
             <div className="row">
                 <div className="col-sm">
-                    
+
                 </div>
                 <div className="col-sm">
                     Your In the Lobby...
```
Vlek: *You're

frontend/src/server/Room.js
```diff
@@ -0,0 +1,33 @@
...
+var io = require("socket.io")(http);
+
+const port = 3000;
+
+app.get("/", function(req, res) {
+  res.sendFile(__dirname + "/index.html");
+});
+
+io.on("connection", function(socket) {
+  console.log("a user connected");
```
Vlek: I would strongly suggest adding a bit more information about the user if possible, potentially like their IP address to be able to differentiate between players.

It's also the case that we're going to have to deal with connecting, disconnecting, and reconnecting. We're going to need to handle cookies or some other way of allowing players to return to an existing game state.

frontend/src/server/Room.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+var app = require("express")();
+var http = require("http").Server(app);
+var io = require("socket.io")(http);
+
+const port = 3000;
+
+app.get("/", function(req, res) {
+  res.sendFile(__dirname + "/index.html");
```
Vlek: Is this using react? I believe this is instead using express, which is a different framework.

frontend/src/server/Room.js
```diff
@@ -0,0 +1,33 @@
...
+
+  socket.on("join_room", room => {
+    socket.join(room);
+  });
+
+    http.listen(port, function() {
+  console.log(`listening on *:${port}`);
+});
+
+socket.emit("new_visitor", visitor);
```
Vlek: It's a bit hard to follow with the indentation, but is this just emitting without there having necessarily been a new visitor?

frontend/src/server/Room.js
```diff
@@ -0,0 +1,33 @@
...
+
+socket.on("visitors", visitors => {
```
Vlek: Is this preferable to having a single visitor socket method and just deal with multiple events? The server could send each separately.

frontend/src/server/index.js
```diff
@@ -0,0 +1,39 @@
+var app = require("express")();
```
Vlek: Now that I am looking at this, this doesn't seem right. You're using a server-side JS engine for presenting websites when in reality the MT server-side should be offering the html files. The client itself should be creating the websocket to the MT server, not to itself and THEN to the MT server.

CameronBaffuto: I understand now, thank you for the feedback. I will cancel this pull request

PR #18, opened 2019-10-20
Inital resolving of dice to victory points

gpfaia:


backend/src/game/turn.py
```diff
@@ -0,0 +1,33 @@
...
+    return collections.counter(dice)
+
+
+def resolve_victory_points_dice(dice_counter):
+    victory_points = 0
+    for die, die_count in dice_counter.items():
+        if die in [DieValue.ONE, DieValue.TWO, DieValue.THREE]:
+            leftover_dice = die_count - 3
+            if leftover_dice >= 0:
+                victory_points += (die.value + leftover_dice)
```
borisv13: I don't like the if in an if in a for. But I also can't think of a better solution right now.

borisv13:
Make sure to get the changes in master in to this branch, should fix the Sonar complaint.

backend/src/game/turn.py
```diff
@@ -0,0 +1,33 @@
...
+from backend.src.game import constants
+from backend.src.game.locations import Locations
+from backend.src.game.dice import DieValue
+
+
+def get_dice_count(dice):
+    return collections.counter(dice)
+
+
+def resolve_victory_points_dice(dice_counter):
```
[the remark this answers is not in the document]
gpfaia: Yeah, was taking the name from the docs, I think calculate is better

backend/src/game/turn.py
gpfaia: Found a way to reduce the if statements, will upload shortly.

backend/src/game/turn.py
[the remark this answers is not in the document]
borisv13: The name will still exist, just in some other class that leverages this method.  Somewhere, we get the number this returns and actually affect the player.

backend/test/game/test_turn.py
```diff
@@ -0,0 +1,24 @@
...
+    return collections.Counter([DieValue.TWO, DieValue.TWO, DieValue.TWO,
+                                DieValue.TWO, DieValue.HEAL, DieValue.ATTACK,
+                                DieValue.ENERGY])
+
+
+def test_resolve_victory_points_dice(dice_list):
+    assert turn.resolve_victory_points_dice(dice_list) == 3
+
+
+def test_resolve_energy_dice(dice_list):
```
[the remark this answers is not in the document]
gpfaia: I am removing these tests as they are reptitive and in the end just testing the functionality of counter (aka not needed)

backend/test/game/test_turn.py
```diff
@@ -0,0 +1,30 @@
+import collections
+import pytest
+
+import backend.src.game.turn as turn
+from backend.src.game.dice import DieValue
+
+
+def test_calculate_victory_points_from_dice():
```
borisv13: Please break this out into separate tests. TDD best practices - one assert per test + each test should only fail for one reason

borisv13: I think it's best to normalize that the list of dice will be 6, I don't think it's possible to be less than 6. Having a special case for 7 might be beneficial because I think some cards may allow for that.

gpfaia: I can break these into multiple tests if you think its best but I think we are getting a bit too into the weeds with the testing this. The list count really shouldn't matter as it is independent of the function and there is no difference if there was 6 dice vs 600 dice in a list

borisv13: I think the tests are plenty as is, they should just be separated.
It was just weird to me that some lists had 5, some 6, some 7 elements

gpfaia: It seems excessive to go through all the possible combinations here, I think just having branch coverage and a few different cases works otherwise you could argue there should be 14 tests (1x sub 3, 1x super 6, 3x for 3-6 die rolls

borisv13: I agree... "I think the tests are plenty as is, they should just be separated."

backend/test/game/test_turn.py
```diff
@@ -0,0 +1,54 @@
+import collections
+import pytest
+
+import backend.src.game.turn as turn
+from backend.src.game.dice.dice import DieValue
+
+
+def test_sub_three_victory_point_dice():
```
borisv13: Sorry to be anal, but it's my job yay.
Please make sure test names describe system behavior as well as system input.
For example:
test_less_than_three_victory_point_dice_no_victory_points_awarded

backend/test/game/test_turn.py
```diff
@@ -0,0 +1,54 @@
...
+
+def test_less_than_three_victory_point_dice_no_victory_points_awarded():
+    dice = collections.Counter([DieValue.TWO, DieValue.TWO,
+                                DieValue.HEAL, DieValue.ATTACK,
+                                DieValue.HEAL, DieValue.ATTACK,
+                                ])
+    assert turn.calculate_victory_points_from_dice(dice) == 0
+
+
+def test_three_victory_point_dice_two_victory_points_awarded():
```
borisv13: I feel that the tests are missing the dice value that was rolled, ie
test_three_twos_rolled_two_victory_points_awarded()
By reading the test names alone, it doesn't make sense.

backend/src/game/turn.py
```diff
@@ -0,0 +1,41 @@
...
+    victory_dice = [DieValue.ONE, DieValue.TWO, DieValue.THREE]
+    for victory_die in victory_dice:
+        die_count = dice_counter[victory_die]
+        leftover_dice = die_count - 3
+        if leftover_dice >= 0:
+            victory_points += (victory_die.value + leftover_dice)
+    return victory_points
+
+
+def calculate_energy_from_dice(dice_counter):
```
borisv13: Can you please add 1 simple test for each of these, energy, attack, heal methods? Maybe even remove the energy method for now since we're not handling that yet

PR #20, opened 2019-10-28
Created a script to quickly build the required react files and run

Vlek:
This is tested for Windows 10, but it should also be able to work
on Linux/Mac.

The only thing I'm not 100% on is the fact that I have global
variables in the file itself. The constants file is nested in
the game folder, so this seemed like it should be separate from
that. Maybe a new constants file at the main folder level for
this? Seemed weird to have for only one thing, that's why I
didn't just do it.

### DIFF
--- a/build_and_run_project.py
+++ b/build_and_run_project.py
@@ -1,0 +1,49 @@
+"""
+King of Tokyo game script
+
+This file performs all of the necessary commands in order to build
+the project and run the Django server so that the game can be
+played locally.
+"""
+
+import subprocess
+import sys
+
+WINDOWS_COMMAND_PORTIONS = ["cmd.exe", "/c"]
+
+GO_TO_FRONTEND_FOLDER = ["cd", "frontend"]
+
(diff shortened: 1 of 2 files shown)


borisv13:
Please exclude this file from code coverage. If help is needed, reach out via Slack I'd be happy to help.

borisv13:
I like having the variables here, as they having nothing to do with the business rules of the system like the constants do.

Vlek:
> 
> 
> Please exclude this file from code coverage. If help is needed, reach out via Slack I'd be happy to help.

I did some research in order to do this and I came across something that we may want to consider. Instead of excluding this file path, we may instead want to specifically state our source paths as being both the front end and back end folders. That seems to be the way that the developers intended sonar to be used, please see this documentation: https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/#NarrowingtheFocus-Patterns

borisv13:
Yep.

PR #22, opened 2019-10-31
Added temp UI Dice component integration with REST endpoint

kaubster:
Infrastructure Changes:
-Added Django REST package
-Added CORS support to allow access by React frontend
-Added api/ route for access rest endpoints
-Added backend/src/kot/views_rest.py containing REST endpoint views. views.py imports views_rest.py to avoid additional Django infrastructure. Note: Location of these files may need to be refactored in the future given remaining task "Need Reorg of Backend folder structure"

Gameboard UI Changes:
-Modified Dice model to support UI rolling
-Added Message model for use to send commands

Travis:
-added stages, backend (python) and frontend (react/nodejs)

Added functionality to DiceRoller and DiceBoard 
- display dice received via REST endpoint. End point is mocked. Was not able to integrate with dice.py given Location of views_api.py unable to import dice.py. Can refactor once remaining task "Need Reorg of Backend folder structure" completed.
- display dice
- allow selection of dice
- disables Roll button when rest indicates number of rolls remaining, no more roles permitted results in disabled roll button

Future Tasks (most to least pressing):
**-BIGGY: Need Reorg of Backend folder structure, not currently possible for backend/src/kot files to reference backend/src/game at the moment.**
-add Django Channel support (REST support will remain since Achievements and Card services (for example) make sense as one off requests via REST)
-DiceRoller does not clear selected dice after each roll, it should.
-communication back to server providing selected dice
-communication back to server providing requested re-roll count
-Once Channels available - refactor request for re-roll to use Channel provided websockets

Note: To avoid dup of effort, I do intend to hit the "BIGGY" next.


frontend/src/components/GameConsole/GameConsole.js
```diff
@@ -54,8 +54,7 @@ class GameConsole extends React.Component {
 
         <form className="submit_game_command"
           onSubmit={this.SubmitGameCommand}
-          onChange={this.handleChange}
-          className="submit_game_command">
```
Lakshmii3: className is repeated twice (Line 55 and 58)

Lakshmii3:
Hi Keith,
1. I see that the containers Login, Lobby and GameBoard have props passed in their functions, but aren't actually taking any parameters. Can props be removed as we are just using the components and not passing any parameters to containers?

2. For now, we have mocked the dice end point and able to roll the dice. I understand that backend folder refactoring has to be addressed for sending calls to API for utilizing selected dice, re-rolling remaining and limit roll count logic. From UI point, I am okay with the changes made so far.

frontend/src/components/GameConsole/GameConsole.js
kaubster: removed extra <form ... className

PR #32, opened 2019-11-09
Starting to setup Channels

gpfaia:
Not complete

https://channels.readthedocs.io/en/latest/tutorial/part_1.html


gpfaia:
I think this currently breaks Django in it's current form so will have to edit it more for it to be stable

gpfaia:
This has been completed in another changelist

PR #33, opened 2019-11-10
Project structure refactor

cvennel:
Refactored project structure for backend. 

Important notes:
- New cards became cards and cards became old cards
  - Old cards were not deleted due to being required by deck, will remove in future PR
- Django stuff left in game, not sure if they need to be there or can move


cvennel:
I'm not sold on the title "definitions" for the directory for constants and the location enum.

PR #35, opened 2019-11-10
Hosting react via django runserver

kaubster:
**Now by default, python manage.py runserver will host the production React site when available.**

Python settings are configured to load the production version of our react site located within directory kot\frontend\build.

Updated kot/readme.md adding _Quick Start_ and _Detailed_ sections.
_Quick Start_: Install/Build React using production settings, start python server to host the site
_Detailed_: 
- Options to run with or without virtual environments
- Running: React and Python servers separately which is necessary often when in development. React watch mode is useful when updating UI.
- Testing: React and Python testing commends


backend/game/urls.py
```diff
@@ -1,7 +1,10 @@
...
+from django.urls import path, include
+from django.conf.urls import url, include
+from django.contrib.auth.models import User
 
 from . import views
 
 urlpatterns = [
-    path('dice/', views.views_rest.roll_dice),
-]
+    url(r'^',views.FrontendAppView.as_view()),
```
Vlek: I won't know for certain until this is all merged and done with, but, with those this routes, it looks like it will be easy to add Channels to this.

PR #38, opened 2019-11-10
Refactored turn.py

cvennel:
turn package became turn actions. turn.py became attack. dice resolution logic moved to dice_resolver in dice package.


backend/game/turn_actions/test_attack.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+import game.turn_actions.attack as attack_actions
+from game.player.player import Player
+
+
+@pytest.fixture(autouse=True)
+def player():
```
gpfaia: This is not needed for this test

PR #39, opened 2019-11-10
Heal

cvennel:
Added heal method that limits healing only while not in tokyo


backend/game/turn_actions/test_heal.py
```diff
@@ -0,0 +1,22 @@
...
+from game.player.player import Player
+
+
+@pytest.fixture(autouse=True)
+def player():
+    player: Player = Player()
+    return player
+
+
+def test_heal_only_in_tokyo(player):
```
borisv13: Please split into 2 tests here.

backend/game/cards/discard_cards/victory_point_manipulation_cards/test_heal.py
```diff
@@ -0,0 +1,10 @@
+from game.cards.discard_cards.victory_point_manipulation_cards.heal import Heal
```
gpfaia: These cards should not be under the victory_point_manipulation_cards folder...

There should be a health_manipulation_cards folder.  Personally not a fan of breaking it down this way (if a card does mutiple things, where does it go?) but for right now that is the pattern we are sticking with

backend/game/turn_actions/test_heal.py
```diff
@@ -0,0 +1,22 @@
...
+    player.leave_tokyo()
+    starting_health = player.current_health
+    player.update_health_by(-3)
+    heal.heal_self(player, 1)
+    assert player.current_health == starting_health - 2
+
+    player.move_to_tokyo()
+    heal.heal_self(player, 1)
+    assert player.current_health == starting_health - 2
```
gpfaia: Professor usually doesn't like two asserts in a single test as it makes for confusion, break these into two separate tests so when it fails, we know the condition that caused it to fail.

backend/game/cards/discard_cards/victory_point_manipulation_cards/heal.py
```diff
@@ -0,0 +1,10 @@
+from game.cards.discard_card import DiscardCard
+from game.cards.card import Card
+
+
+class Heal(DiscardCard):
```
gpfaia: We are going to have to be very careful with this card and the turn action heal and making sure we are not overloading or misreferencing things but we can figure that out at a later date.

backend/game/turn_actions/test_heal.py
gpfaia: Also make sure there is a newline at the end of the file

backend/game/turn_actions/heal.py
```diff
@@ -0,0 +1,7 @@
+import game.player.player as player
+import game.values.locations as locations
+
+
+def heal_self(player_healing: player.Player, health_points_to_add):
+    if player_healing.location is not locations.Locations.TOKYO:
+        player_healing.update_health_by(health_points_to_add)
```
borisv13: So we need to be careful here.
The rule is: the player cannot heal _from dice_ while in Tokyo. 
The player _can_ heal from cards or other means while in Tokyo.
We need test cases for all of this as well.

cvennel: My thought was this would only be used from dice resolution (should have named it better) since cards seemed to be using the player.update_health_by() method. That was why I left alone the card part. Turns out the heal card I was referencing wasn't in master but I saw it cause I accidentally made a branch the same name as @otaiwo70 who was working on the heal card. To better reflect my original intention heal_self should be renamed to heal_self_from_dice. But if we're changing the heal card then I need to account for that.

backend/game/cards/discard_cards/victory_point_manipulation_cards/heal.py
cvennel: I think this was actually written by @otaiwo70 and I accidentally pulled it in. Tagged so he sees your comment.

backend/game/cards/discard_cards/victory_point_manipulation_cards/test_heal.py
cvennel: I think this was actually written by @otaiwo70 and I accidentally pulled it in. Tagged so he sees your comment.

PR #40, opened 2019-11-10
Removing old cards

cvennel:
removing old card class. Added methods to search for all card classes and add to deck using:

- backend/game/cards/discard_cards/victory_point_manipulation_cards/card_collector.py
- backend/game/cards/master_card_list


backend/game/cards/discard_cards/victory_point_manipulation_cards/card_collector.py
```diff
@@ -0,0 +1,36 @@
+from os.path import dirname, basename, isfile, join
```
gpfaia: NIT: When using a package like os, I prefer to just import the entire thing.  I know that there is nothing wrong with importing the os.path but I personally like to see the entire package in the name space especially because more often than not, a lot of functions are used in it (and just better readability for me)

backend/game/cards/discard_cards/victory_point_manipulation_cards/apartment_building.py
```diff
@@ -8,3 +8,5 @@ def __init__(self):
 
     def immediate_effect(self, player_that_bought_the_card, other_players):
         player_that_bought_the_card.update_victory_points_by(3)
+
```
gpfaia: Only need 1 space at the end of a file

backend/game/cards/discard_cards/victory_point_manipulation_cards/card_collector.py
```diff
@@ -0,0 +1,36 @@
...
+def get_class(class_string_name_and_path):
+    parsed_class_string = class_string_name_and_path.split('.')
+    module = ".".join(parsed_class_string[:-1])
+    m = __import__(module)
+    for comp in parsed_class_string[1:]:
+        m = getattr(m, comp)
+    return m
+
+
+def get_list_vpm_cards(prepend_path):
```
gpfaia: Initial gut reaction is I do not like this because I don't think it's clean or substainable, I will need to look more into a better example of how to go about this but I think we can utilitize the __init__ files for this purpose.  Yes we might have to list out all the cards but I think that's okay as it'll only need to be done once

https://www.bensnider.com/dynamically-import-and-instantiate-python-classes.html
http://www.pressthered.com/dynamically_importing_and_instantiating_modules_in_python/

gpfaia: Yes after playing around with the `__init__` files a bit, you can probably get them to properly call the subclasses here so you do not have to dynamically find them.  In the end, when you import cards, basically it makes sense for all cards to be imported.  I think that may be better to try before coming up with this sort of dynamic function

borisv13: I agree. Also there was multiples of some cards, so I'd like the deck to be listed out manually as well.

cvennel:
@gpfaia  @borisv13 Thanks for taking a look, wanted to challenge myself with an experiment and see if I could come up with a creative solution. I'll scrap that part and go to a manual list.

PR #44, opened 2019-11-11
Jet fighters

otaiwo70:


backend/game/cards/discard_cards/multi_manipulation_cards/test_jet_fighters.py
```diff
@@ -0,0 +1,16 @@
...
+
+def test_jet_fighters_adds_5_victory_points(player):
+    JetFighters().immediate_effect(player, None)
+    assert player.victory_points == 5
+
+
+def test_jet_fighters_subtracts_4_health(player):
+    player.current_health = 9
+    JetFighters().immediate_effect(player, None)
+    assert player.current_health == 5
```
gpfaia: You could do this as an "and" statement of checking both simultaneously due to the nature of the card but not required

otaiwo70: player.current_health = 9 removed

PR #45, opened 2019-11-11
National guard

otaiwo70:


backend/game/cards/discard_cards/multi_manipulation_cards/test_national_guard.py
```diff
@@ -0,0 +1,16 @@
+from game.cards.discard_cards.multi_manipulation_cards.national_guard import NationalGuard
+
+
+def test_jet_fighters_adds_2_victory_points(player):
+    NationalGuard().immediate_effect(player, None)
+    assert player.victory_points == 2
+
+
+def test_jet_fighters_subtracts_2_health(player):
+    player.current_health = 8
```
gpfaia: No need to set the default health down

otaiwo70: pre set removed

borisv13: Why 😭

PR #46, opened 2019-11-11
Nuclear power plant

otaiwo70:


backend/game/cards/discard_cards/multi_manipulation_cards/test_nuclear_power_plant.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_jet_fighters_adds_2_victory_points(player):
+    NuclearPowerPlant().immediate_effect(player, None)
+    assert player.victory_points == 2
+
+
+def test_jet_fighters_adds_3_health(player):
+    NuclearPowerPlant().immediate_effect(player, None)
+    assert player.current_health == constants.DEFAULT_HEALTH
```
borisv13: This is not a good test. You are not asserting that the health went up by 1. You are testing that the max cap size works.
Either damage the player as part of setup, or set the player's health as part of setup. 7 would be a good number, since then the assert can be DEFAULT_HEALTH

PR #49, opened 2019-11-11
Heal turn methods

cvennel:


backend/game/turn_actions/heal.py
```diff
@@ -0,0 +1,11 @@
+from game.player.player import Player
+import game.values.locations as locations
+
+
+def heal_self_from_dice(player_being_healed: Player, health_points_to_add):
+    if player_being_healed.location is not locations.Locations.TOKYO:
```
gpfaia: For readability, I tend to avoid using not whenever possible (just an extra evaluation I have to do in my head).  This can just be changed to equals locations.Locations.OUTSIDE

Also may want to verify that `is` is the correct operator here because could make a case for ==

backend/game/turn_actions/test_heal.py
```diff
@@ -0,0 +1,38 @@
...
+
+@pytest.fixture(autouse=True)
+def hurt_player():
+    hurt_player: Player = Player()
+    hurt_player.update_health_by(-3)
+    return hurt_player
+
+
+def test_heal_from_dice_out_of_tokyo(hurt_player):
+    hurt_player.leave_tokyo()
```
borisv13: Maybe I'm being pedantic, but I don't like .leave_tokyo() here, since that's a verb, an actions, to be triggered when the player is leaving tokyo.
Instead, can we do hurt_player.location == Locations.OUTSIDE? This way it's clear that you're just setting the location, not talking about actually moving locations.

backend/game/turn_actions/test_heal.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_no_heal_from_dice_in_tokyo(hurt_player):
+    hurt_player.move_to_tokyo()
+    starting_health = hurt_player.current_health
+    heal.heal_self_from_dice(hurt_player, 1)
+    assert hurt_player.current_health == starting_health
+
+
+def test_heal_from_dice_out_of_tokyo(hurt_player):
```
borisv13: I'm confused by the last 2 tests

backend/game/turn_actions/heal.py
borisv13: I agree. Also when we add Tokyo Bay, the logic becomes much easier to check outside, vs checking both Tokyos

backend/game/turn_actions/heal.py
```diff
@@ -0,0 +1,11 @@
...
+        player_being_healed.update_health_by(health_points_to_add)
+
+
+def heal_self_no_tokyo_restriction(player_being_healed: Player, health_points_to_add):
```
borisv13: I don't think this method needs to exist?
So far, cards that heal call update_health_by() themselves

backend/game/turn_actions/heal.py
cvennel: Ooh I completely missed the difference between is and == thanks. Will update.

backend/game/turn_actions/heal.py
cvennel: I didn't think it needed to exist either. Then I second guessed myself and thought I remembered being told we needed it there for some reason I couldn't remember. I'd rather remove it too. will delete

backend/game/turn_actions/test_heal.py
cvennel: It would help if I named it something that made any sort of sense. Will be deleted anyways with removing no_tokyo_restriction method

backend/game/turn_actions/test_heal.py
cvennel: Ok I can change that

backend/game/turn_actions/test_heal.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@pytest.fixture(autouse=True)
+def hurt_player():
+    hurt_player: Player = Player()
+    hurt_player.update_health_by(-3)
+    return hurt_player
+
+
+def test_heal_from_dice_out_of_tokyo(hurt_player):
```
borisv13: I'd like to see both of these tests exist PLUS:
The same 2 tests, but actually hitting the dice resolver method, to assert that the dice resolver uses this methods and behaves correctly.

PR #50, opened 2019-11-12
Channels added!

Vlek:
Just accept! Don't even read this, Accept!


kaubster:
good starting point, UI can attempt to connect with socket.

PR #51, opened 2019-11-16
Heal

otaiwo70:


backend/game/cards/discard_cards/health_manipulation_cards/test_heal.py
```diff
@@ -0,0 +1,11 @@
+from game.cards.discard_cards.health_manipulation_cards.heal import Heal
+from game.values import constants
+
+
+def test_heal_adds_2_health(player):
+    Heal().immediate_effect(player, None)
+    assert player.current_health == constants.DEFAULT_HEALTH
```
gpfaia: Technically this is correct in that the assert will work if the player is at max/default health it will not add 2 health but this does not test if the player is below the default if it will actually add the two

backend/game/cards/discard_cards/multi_manipulation_cards/test_gas_refinery.py
```diff
@@ -14,4 +14,4 @@ def test_gas_refinery_adds_2_victory_points_to_current_player(player, five_playe
 def test_gas_refinery_subtracts_3_health_from_other_players(player, five_players):
     GasRefinery().immediate_effect(player, five_players)
     assert all(other_players.current_health ==
-               7 for other_players in five_players)
\ No newline at end of file
+               7 for other_players in five_players)
```
gpfaia: This should be `constants.DEFAULT_HEALTH - 3` instead of the int 7

backend/game/cards/discard_cards/health_manipulation_cards/test_heal.py
otaiwo70: change made and pushed to recent commit

backend/game/cards/discard_cards/multi_manipulation_cards/test_gas_refinery.py
otaiwo70: change made and pushed to recent commit

backend/game/cards/discard_cards/health_manipulation_cards/test_heal.py
gpfaia: Again, technically correct but does suffer some testing flaws.  Should
1. Probably best to use the method in player that adjust the health by a set value
2.  If are going to adjust the health manually need to set it to `player.current_health = constants.DEFAULT_HEALTH - 2` but I recommend 1 personally

otaiwo70:
Made changes to code upon talks with the Professor, needs to be merged if all issues are met

backend/game/cards/discard_cards/health_manipulation_cards/test_heal.py
gpfaia: The above comment has not been addressed

PR #54, opened 2019-11-16
First pass of adding player queue

gpfaia:
Spent some time reading through the rules set, going through and testing different state machine styles, and looking for easy code without forcing bad practices.

I think this design pattern will work but wanted to put it in front of everyone before committing more towards it and adding it to a game board functionality


backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,36 @@
...
+    def _add_initial_player(self, player):
+        self.all_players.append(player)
+
+    def _remove_initial_player(self, player):
+        if player in self.all_players:
+            self.all_players.remove(player)
+
+    def set_player_order(self):
+        # TODO: set this based off of dice roll minigame
+        # Currently just the order in which they were added to the list
```
borisv13: Not a priority TODO, keep this order for a long while

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,36 @@
+import itertools
+
+
+class GamePlayers():
+
+    def __init__(self):
+        self.all_players = self.dead_players = self.alive_players = []
```
borisv13: reference issue here?

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,36 @@
...
+        self.current_player = None
+
+    def _add_initial_player(self, player):
+        self.all_players.append(player)
+
+    def _remove_initial_player(self, player):
```
borisv13: What does this mean? Remove initial player?

gpfaia: Depending on the mechanics of the game lobby, didn't know if players were going to be able to queue in/out so that they could be added to the list and removed but then once a 'start game' (or in this case 'set player order') was established, they would be locked in

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,38 @@
...
+            self.all_players.remove(player)
+
+    def set_player_order(self):
+        # TODO: set this based off of dice roll minigame
+        # Currently just the order in which they were added to the list
+        self.alive_players = self.all_players
+        self.player_cycle = itertools.cycle(self.all_players)
+
+    def get_current_player(self):
+        if not self.current_player:
```
samridhprasad: I don't quite understand this condition before returning the current player?

gpfaia: This checks for a null, think on startup, so what happens is it’ll then cycle the next player which will be realistically the current player. If you called this on the next turn around, you would get the real current player as we don’t need to go through the cycle

backend/game/engine/board.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self):
+        self.players = GamePlayers()
+        self.STATUS = status.SETUP
+        self.WINNER = None
+
+    def start_game(self):
+        self.players.set_player_order()
+        self.STATUS = status.ACTIVE
+
+    def check_if_winner(self, player):
```
borisv13: You can also win by being last player alive

backend/game/engine/board.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def end_game(self):
+        self.STATUS = status.COMPLETED
+
+    def turn(self, current_player):
+        # roll dice & dice_resoultion method
+        pass
+
+    def play_game(self):
+        while self.STATUS == status.ACTIVE:
```
borisv13: Considering making this a method for readability, like is_game_active()

backend/game/engine/board.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def end_game(self):
+        self.STATUS = status.COMPLETED
+
+    def turn(self, current_player):
+        # roll dice & dice_resoultion method
+        pass
+
+    def is_game_active(self):
+        if self.STATUS == status.ACTIVE:
```
Vlek: You could instead just `return self.STATUS == status.ACTIVE` since that expression results in a boolean value.

backend/game/engine/board.py
```diff
@@ -0,0 +1,43 @@
...
+            return True
+        else:
+            return False
+
+    def play_game(self):
+        while self.is_game_active():
```
Vlek: This is going to cause the Django server to potentially freeze up. While this is how you would do something like a command line version, there's other code that has to deal with reading from the sockets, handling connections, and serving the website that must regain control at some point.

I do not believe that this method is necessary. Instead, `turn` should probably check whether the player that is attempting to have a turn is the current player and throw an error otherwise. the winner check should probably be stashed in the end-of-turn event code.

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
+import itertools
+
+
+class GamePlayers():
+
+    def __init__(self):
+        self.all_players = []
+        self.dead_players = []
```
Vlek: You don't need to keep lists of dead and alive players. Instead, you can iterate through and pass the players that are dead. It does not make sense to keep multiple lists as they can get out of sync very quickly. That's a lot of headache for no real upsides.

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
...
+        self.alive_players = []
+        self.current_player = None
+
+    def _add_initial_player(self, player):
```
Vlek: Why is this called add_initial_player? Are there going to be other kinds of players that could be added later on in the game?

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
...
+        self.all_players.append(player)
```
Vlek: You're appending whatever is given, but there really should be a player object being passed in. If it isn't a player object, an error should be raised.

backend/game/engine/player_queue.py
Vlek: This really depends on how we have the lobby set up. We could have it such that there's like a chatroom-esque pre-game lobby deal where players can figure out if they want to play. Once they all pick their monsters and what not, then they could be redirected to the actual game playing display where they are assigned a playing order and spots.

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _remove_initial_player(self, player):
+        if player in self.all_players:
```
Vlek: If it's the case that they're not in there, then it should throw an error.

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_current_player(self):
+        if not self.current_player:
+            self.get_next_player()
+        return self.current_player
+
+    def get_next_player(self):
+        self.current_player = self._cycle_next_alive_player()
+
+    def get_alive_player(self):
```
Vlek: I think this was supposed to be players plural.

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
...
+        return self.alive_players
```
Vlek: As stated earlier, this should probably look through all players and simply recheck whether they're alive. A helper `_is_alive` method would probably be best.

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_all_other_alive_player(self):
+        return [player for player in self.alive_players if
+                self.current_player != player]
+
+    def _cycle_next_alive_player(self):
+        for player in self.player_cycle:
```
Vlek: Where is this player_cycle? Why wouldn't you just use `self.players`?

backend/game/engine/player_queue.py
```diff
@@ -0,0 +1,45 @@
...
+            if player in self.alive_players:
+                return player
+            elif not self.alive_players:
+                return None
```
Vlek: This makes your for-loop unnecessary, you're just returning the first player or None if they're alive. You should cycle through and then have this at the end as an error if there are no alive players potentially. However, this should never happen because the last one standing would be considered the winner.

backend/game/engine/player_queue.py
gpfaia: The cycle is in line 31, once the game begins and the player list/order is set, a cycle is created so that we do not need to worry about indexes or positions or dealing with the list type in a way that goes against python best pratices, we can just iterate through the cycle.

backend/game/engine/player_queue.py
gpfaia: The cycle will keep state of the last position so that is untrue.  I have provided an example of what I mean.  Now I realize that I am passing in the cycle object each time but since the cycle is in the self of the class it is the equivalent behavior.

>>> import itertools
>>> a = ['foo', 'bar', 7, 2]
>>> b = itertools.cycle(a)
>>> def cycle_b(cycle_obj):
...  for x in cycle_obj:
...   return x
...
>>> cycle_b(b)
'foo'
>>> cycle_b(b)
'bar'
>>> cycle_b(b)
7
>>> cycle_b(b)
2
>>> cycle_b(b)
'foo'
`

backend/game/engine/player_queue.py
gpfaia: https://stackoverflow.com/a/32684829

and type annotations are not enforcable in python, do not want others to assume that python is acting as a statically typed language when they do use types
https://stackoverflow.com/a/32684754

backend/game/engine/board.py
```diff
@@ -16,28 +16,26 @@ def start_game(self):
...
-                in self.players.get_alive_player()):
-            self.WINNER = player
+    def check_if_winner(self, potential_winner):
+        if (potential_winner.victory_points == constants.VICTORY_POINTS_TO_WIN or
+                players.if_last_player_alive(potential_winner)):
+            self.WINNER = potential_winner
             self.end_game()
 
     def end_game(self):
         self.STATUS = status.COMPLETED
```
Vlek: both `self.STATUS` as well as `self.WINNER` should be lower case as they are not object-level constants.

backend/game/engine/board.py
```diff
@@ -16,28 +16,26 @@ def start_game(self):
...
 
-    def turn(self, current_player):
-        # roll dice & dice_resoultion method
-        pass
+    def turn(self, active_player):
+        if active_player == self.players.get_current_player():
```
Vlek: This isn't pythonic, the convention would be if not blah blah blah, throw exception. Forget the else case, just unindent the code and carry on.

backend/game/engine/board.py
```diff
@@ -16,28 +16,26 @@ def start_game(self):
...
+            self.turn_actions(active_player)
+            self.check_if_winner(active_player)
+        else:
+            # TODO player object should have a name property here
```
Vlek: We aren't supposed to be using comments.

backend/game/engine/board.py
```diff
@@ -16,28 +16,26 @@ def start_game(self):
...
-            return False
+        return self.STATUS == status.ACTIVE
 
-    def play_game(self):
-        while self.is_game_active():
-            current_player = self.players.get_current_player()
-            self.turn(current_player)
-            self.check_if_winner(current_player)
+    def turn_actions(self, turn_player):
+        # roll dice & dice_resoultion method
```
Vlek: The prof. dinged me for this. We are not supposed to be adding comments.

backend/game/engine/player_queue.py
```diff
@@ -1,25 +1,21 @@
...
         self.current_player = None
 
-    def _add_initial_player(self, player):
-        self.all_players.append(player)
-
-    def _remove_initial_player(self, player):
-        if player in self.all_players:
-            self.all_players.remove(player)
+    def _add_player_to_game(self, new_player):
+        if isinstance(new_player, Player):
```
Vlek: Again, not pythonic. If it's not an instance of that thing, then raise an error. Otherwise, move on unindented.

backend/game/engine/player_queue.py
```diff
@@ -1,25 +1,21 @@
 import itertools
 
+from game.player.player import Player
+
 
 class GamePlayers():
 
     def __init__(self):
         self.all_players = []
```
Vlek: `self.all_players` makes it seem like you have other lists which do not contain all the players. This could probably be shortened to just `self.players`

backend/game/engine/player_queue.py
Vlek: This didn't make sense because you're checking whether there are any alive players at every iteration unnecessarily. That should probably be a condition before your for-loop where you raise an exception if there aren't any alive players.

backend/game/engine/board.py
gpfaia: Sure but I do want to be careful with how we handle these exceptions before we go down the dangerous rabbit hole of creating custom exceptions for everything.

backend/game/engine/board.py
Vlek: Gerardo resolved this issue by adding this logic.

backend/game/engine/board.py
Vlek: He has also made this method as requested.

PR #57, opened 2019-11-16
Even Bigger

otaiwo70:
added methods to support storage of keep cards to player in dict

Ask : Would like to rename abstract method class DiscardCard(Card): otherwise will need to have a similar method for class KeepCard(Card):

Next steps: class KeepCard(Card): would facilitate storage of keep cards to player dict card store for now, at least i can then start working on templates of all the keep cards, can do that much for all cards before working out additional logic for each card

let me have your feedback on tentative next steps and ask


samridhprasad:
- We need to follow the same structure as the `discard_cards` directory where all the keep cards can be categorized under `keep_cards`
- As discussed with @borisv13 we need to have an abstract method under discard_card.py that implements the functionality for keep cards. 
- Touching Player.py with keep cards logic will get messy, we should move that away

otaiwo70:
true that was my initial approach but it seemed repetitive of sorts, especially in reference to the class definitions. Now have keep_cards folder of card declaration structured similar to discard_cards but the cards need to be kept and connected to a specific player thus in addition to the declaration of the card they need to be kept in that players card store, this is why i added the changes to player. However, this is starting template but the logic is so intertwined and has a tendency to get messy that it will need to be discussed further for best approach, look forward to hearing from all

backend/game/cards/keep_cards/dice_roll_manipulation_cards/acid_attack.py
```diff
@@ -0,0 +1,14 @@
...
+
+class AcidAttack(KeepCard):
+    def __init__(self):
+        Card.__init__(self, "Acid Attack", 6, "Add [attack] to your roll", "")
+
+    def immediate_effect(self, player_that_bought_the_card, other_players):
+        Card.name = "Acid Attack"
+        Card.cost = 6
+        Card.footnote = ""
+        Card.effect = "Add [attack] to your roll"
```
borisv13: Why do lines 10-13 exist? Line 7 has all of this...

backend/game/cards/keep_cards/dice_roll_manipulation_cards/test_acid_attack.py
```diff
@@ -0,0 +1,10 @@
+from game.cards.keep_cards.dice_roll_manipulation_cards.acid_attack import AcidAttack
+
+
+def test_acid_attack_player_add_card(player):
+    AcidAttack().immediate_effect(player, None)
+    assert player.card_on_hand.get("Acid Attack") == "Add [attack] to your roll"
```
borisv13: You're going to write this test for every single keep card in the entire deck?
OR you can put the "add this card to the player" into KeepCard and test it once.

backend/game/player/test_player.py
```diff
@@ -93,3 +94,22 @@ def test_player_energy_cannot_go_below_zero(player):
...
+    Card.effect = "Add [attack] to your roll"
+    player.add_keep_card(Card)
+    assert player.card_on_hand.get("Acid Attack") == "Add [attack] to your roll"
+
+
+def test_player_remove_keep_card(player):
+    Card.name = "Acid Attack"
+    Card.cost = 6
+    Card.footnote = ""
+    Card.effect = "Add [attack] to your roll"
```
borisv13: But why

borisv13: card = AcidAttack()
player.add_keep_card(card)
player.remove_keep_card(card)

backend/game/cards/keep_card.py
```diff
@@ -0,0 +1,9 @@
+from abc import abstractmethod
+
+from game.cards.card import Card
+
+
+class KeepCard(Card):
+    @abstractmethod
```
gpfaia: This abstract method should be placed on the Card class and taken out of the Keep/Discard classes

backend/game/player/player.py
```diff
@@ -2,13 +2,17 @@
...
+
+
 class Player:
     def __init__(self):
         self.maximum_health = self.current_health = constants.DEFAULT_HEALTH
         self.location = Locations.OUTSIDE
         self.is_alive = True
         self.victory_points = constants.DEATH_HIT_POINT
         self.energy = constants.DEFAULT_ENERGY_CUBE
+        self.card_on_hand = {}
```
gpfaia: Dictionary assumes that there is going to be one of each card as the key, this is not true as the mimic card alone will allow mutiple copies of the same card to exist.  This data structure will need to be reworked, my suggestion would be a list

backend/game/player/player.py
```diff
@@ -32,3 +36,9 @@ def update_energy_by(self, change_integer):
         self.energy += change_integer
         if self.energy < constants.DEFAULT_ENERGY_CUBE:
             self.energy = constants.DEFAULT_ENERGY_CUBE
+
+    def add_keep_card(self, card):
+        self.card_on_hand[card.name] = card.effect
```
gpfaia: I do not see how this is helpful or necessary, the effect is a string representation of what the card should do,

backend/game/player/player.py
```diff
@@ -32,3 +36,9 @@ def update_energy_by(self, change_integer):
...
+
+    def remove_keep_card(self, card):
+        self.card_on_hand.pop(card.name)
```
gpfaia: If changing to a list, this method will have to be redone

cvennel: This is redone in #68

cvennel:
closes #83 
closes #100

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
```diff
@@ -0,0 +1,17 @@
+from game.cards.card import Card
+from game.cards.keep_card import KeepCard
+
+
+class EvenBigger(KeepCard):
+    def __init__(self):
+        Card.__init__(self, "Even Bigger", 4,
+                      "+2[health] when you buy this card.maximum [health] increased to 12 as long as you own this card",
```
gpfaia: This card will need to have the effect of losing the maximum health set back to the original value but this could either be done now or in a later iteration

cvennel:
Outdated, superseded by other PRs

PR #61, opened 2019-11-20
Eater of the dead

samridhprasad:
- Implemented and tested victory point manipulation keep card
- Restructured card.py, discard_card.py, keep_card.py to have appropriate abstract methods
- Added support for holding keep cards in player.py
- Unit tested player.py and eater_of_the_dead.py


backend/game/cards/keep_cards/victory_points_manipulation_cards/eater_of_the_dead.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class EaterOfTheDead(KeepCard):
+    def __init__(self):
+        Card.__init__(self, "Eater of the Dead", 4, "+ 3[Star] every time a monster's [Heart] goes to 0.")
+
+    def special_effect(self, player_that_bought_the_card, other_players):
+        for other_player in other_players:
+            if other_player.current_health == constants.DEATH_HIT_POINT:
+                player_that_bought_the_card.update_victory_points_by(3)
```
borisv13: What triggers special_effect?
I don't think this code works.
When player 1 dies, this special effect kicks off and gains them 3 VP
When player 2 dies, this special effect kicks off and gains them 6 VP, because 2 players are now dead.

I believe this for/if doesn't work, and instead this method should just be "increase VP by 3" and then the hard part is only calling this method when a player dies.

backend/game/cards/discard_card.py
```diff
@@ -4,6 +4,5 @@
 
 
 class DiscardCard(Card):
-    @abstractmethod
     def immediate_effect(self, player_that_bought_the_card, other_players):
         raise NotImplementedError
```
gpfaia: This whole function can be removed as it is no longer needed here, it may also be possible to roll up all this discard/keep card modules into the Cards module.

backend/game/cards/keep_card.py
```diff
@@ -0,0 +1,11 @@
+from abc import abstractmethod
+from game.cards.card import Card
+
+
+class KeepCard(Card):
+    def immediate_effect(self, player_that_bought_the_card, other_players):
+        player_that_bought_the_card.add_card(self)
```
gpfaia: Remove from class

backend/game/cards/keep_card.py
```diff
@@ -0,0 +1,11 @@
...
+
+    @abstractmethod
+    def special_effect(self, player_that_bought_the_card, other_players):
```
gpfaia: I think this should be called maybe like passive effect, but the issue i can see here is that it can also be determined by different things (such as dice rolls, etc) takes in more than just the player who bough card and other players

backend/game/cards/keep_cards/victory_points_manipulation_cards/eater_of_the_dead.py
gpfaia: Maybe these cards are responsible for their own state?  For example, this one could keep a list of the players who died.  Yes there may be logic that might be shared across cards but we can factor them out over time if we treat each individual one as it own entity?  Just a possible thought, curious on @borisv13 opinion in regards to this

backend/game/player/player.py
```diff
@@ -32,3 +37,15 @@ def update_energy_by(self, change_integer):
...
         if self.energy < constants.DEFAULT_ENERGY_CUBE:
             self.energy = constants.DEFAULT_ENERGY_CUBE
+
+    def add_card(self, card):
+        self.card_on_hand.append(card)
+
+    def remove_card(self, index):
+        self.card_on_hand.pop(index)
+
+    def get_card_at_index(self, index):
```
gpfaia: I would prefer a different access style then doing everything by numeric index, think about how this is being utilized.  You're not going to be saying get_card_at_index(0), the player is most likely to say "get_keep_card(KEEP_CARD_NAME)"

backend/game/player/player.py
```diff
@@ -32,3 +37,15 @@ def update_energy_by(self, change_integer):
         self.energy += change_integer
         if self.energy < constants.DEFAULT_ENERGY_CUBE:
             self.energy = constants.DEFAULT_ENERGY_CUBE
+
+    def add_card(self, card):
+        self.card_on_hand.append(card)
+
+    def remove_card(self, index):
+        self.card_on_hand.pop(index)
```
gpfaia: Pop will also act as return as an FYI

cvennel:
closes #96

PR #66, opened 2019-11-25
attached player to energy resolution in dice_resolver

cvennel:
Small change to make dice_resolver automatically update player energy

### DIFF
--- a/backend/game/dice/dice_resolver.py
+++ b/backend/game/dice/dice_resolver.py
@@ -37,7 +37,10 @@ def dice_resolution(dice, player):
     health = calculate_heal_from_dice(dice_counter)
     heal_self_from_dice(player, health)
     attack = calculate_attack_from_dice(dice_counter)
+
     energy = calculate_energy_from_dice(dice_counter)
+    player.update_energy_by(energy)
+
     return victory_points, health, attack, energy
 
 
--- a/backend/game/dice/test_dice_resolver.py
+++ b/backend/game/dice/test_dice_resolver.py
@@ -2,6 +2,8 @@ import collections
 
 from game.dice.dice import DieValue
 import game.dice.dice_resolver as dice_resolver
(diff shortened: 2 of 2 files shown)


backend/game/dice/dice_resolver.py
```diff
@@ -37,7 +37,10 @@ def dice_resolution(dice, player):
     health = calculate_heal_from_dice(dice_counter)
     heal_self_from_dice(player, health)
     attack = calculate_attack_from_dice(dice_counter)
+
     energy = calculate_energy_from_dice(dice_counter)
+    player.update_energy_by(energy)
```
gpfaia: Not a huge fan of doing all these resolutions and returning there values in one method but functionally it all works

cvennel: I'm not either. I think whoever started this had another idea in mind and it has kind of morphed to current state. I wasn't sure if I should just remove the return statement or not

cvennel: I'm not either. I think whoever started this had another idea in mind and it has kind of morphed to current state. I wasn't sure if I should just remove the return statement or not

cvennel: I'm not either. I think whoever started this had another idea in mind and it has kind of morphed to current state. I wasn't sure if I should just remove the return statement or not

gpfaia: Guilty of being the original author.  I was doing that because the game technically allows you to resolve the dice in any order so that method would find all the values and then I was going to create like energy_turn_resolve, attack_turn_resolve, etc. and pass in the value that was calculated for each element in the previous function to those seperate functions.  For round 1 of the game, this realistically shouldnt matter if all done in 1 method BUT once cards are added, there are benefits for doing one before another (example may one to heal before attacking because don't want to move into toyko and not be able to heal)

PR #68, opened 2019-12-01
Made cards cost energy, added player list of cards

cvennel:
Simple edits to make purchasing cards actually cost energy

NOTE: This implements the List change that Gerardo suggested in #57 review notes which could lead to some conflicting work.

### DIFF
--- a/backend/game/deck/deck_handler.py
+++ b/backend/game/deck/deck_handler.py
@@ -3,6 +3,7 @@ from typing import List
 import game.values.constants as constants
 from game.cards.card import Card
 from game.deck.deck import Deck
+from game.player.player import Player
 
 
 class DeckHandler:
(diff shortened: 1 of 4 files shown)


backend/game/deck/deck_handler.py
```diff
@@ -58,10 +59,15 @@ def shuffle_discard_pile_to_draw_pile(self):
             self.__draw_pile.append(self.discard_pile.draw_from())
         self.__discard_pile.clear()
 
-    def buy_card_from_store(self, index):
-        bought = self.__card_store.pop(index)
-        self.__fill_card_store()
-        return bought
+    def buy_card_from_store(self, index, purchasing_player: Player):
+        card_to_buy: Card = self.__card_store[index]
+        if purchasing_player.energy < card_to_buy.cost:
```
gpfaia: Stated this in a similar review, maybe best to have this check sit outside this method, but could see an argument for either way

PR #69, opened 2019-12-01
added sweep card store method

cvennel:


backend/game/deck/deck_handler.py
```diff
@@ -68,3 +69,12 @@ def discard(self, card: Card, card_from_location: List[Card] = None):
         self.__discard_pile.append(card)
         if card_from_location is not None:
             card_from_location.remove(card)
+
+    def sweep_store(self, player_invoking_sweep: Player):
+        if player_invoking_sweep.energy < constants.SWEEP_CARD_STORE_COST:
```
gpfaia: I wonder if it is better to have this check outside of the sweep method as it would make it easier to allow an endpoint to check if a sweep is possible and also removes some of the complexity.

cvennel:
@gpfaia I just resolved a merge conflict with the constants file which dismissed your review, mind re-approving?

backend/game/deck/deck_handler.py
cvennel: I merged the other PR that had your similar comment. I think if we find a better spot to perform those checks then we can open a new PR to move it in the future to support the endpoint but I think for now this works

PR #70, opened 2019-12-02
Solar powered keep card

otaiwo70:


backend/game/cards/keep_cards/energy_manipulation_cards/solar_powered.py
```diff
@@ -0,0 +1,18 @@
...
+class SolarPowered(KeepCard):
+    def __init__(self):
+        Card.__init__(self, "Solar Powered", 4, "At the end of your turn gain 1[Energy] if you have no [Energy].", "")
+
+    def immediate_effect(self, player_that_bought_the_card, other_players):
+        card = SolarPowered()
+        player_that_bought_the_card.add_card(card)
+
+    def special_effect(self, player_that_bought_the_card, other_players):
+        if player_that_bought_the_card.energy == constants.DEATH_HIT_POINT:
```
borisv13: Create a new constant, reading this line makes no sense

cvennel:
closes #138

backend/game/cards/keep_card.py
```diff
@@ -0,0 +1,13 @@
+from abc import abstractmethod
+
+from game.cards.card import Card
```
cvennel: This file should be replaced with the keep card class that is already on master

backend/game/cards/keep_cards/energy_manipulation_cards/solar_powered.py
cvennel: Constant for 0 is gone

backend/game/cards/keep_cards/energy_manipulation_cards/test_solar_powered.py
```diff
@@ -0,0 +1,18 @@
...
+def test_solar_powered_player_add_card(player: Player):
+    SolarPowered().immediate_effect(player, None)
+    assert player.has_instance_of_card(SolarPowered())
+
+
+def test_solar_powered_costs_4_energy():
+    assert SolarPowered().cost == 4
+
+
+def test_solar_powered_gain_1_star_when_energy_is_0(player):
```
cvennel: This is testing the wrong thing

cvennel: oops read it wrong

backend/game/engine/board.py
```diff
@@ -29,6 +30,8 @@ def turn(self, active_player):
         if active_player != self.players.get_current_player():
             raise Exception("It is not %s's turn" % (id(active_player)))
         self.turn_actions(active_player)
+        if active_player.has_instance_of_card(SolarPowered()):
```
gpfaia: Although this is where the card would be active, I think writing a separate function for all end of turn cards would be best for now as I can see this part of the code expanding or us taking it in a different direction so would want it to be slightly separated from the main turn function

cvennel: I was just thinking about that. I think that should be a future refactor since these requests have just lived way too long and this can be good enough for now.

PR #71, opened 2019-12-03
Initial commit for it has a child keep card

otaiwo70:


backend/game/player/player.py
```diff
@@ -32,3 +33,21 @@ def update_energy_by(self, change_integer):
...
+    def get_card_at_index(self, index):
+        return self.card_on_hand.pop(index)
+
+    def get_current_hand(self):
+        return self.card_on_hand
+
+    def discard_all_cards(self):
+        self.card_on_hand.clear()
+
+    def lose_all_stars(self):
```
gpfaia: This function is a bit unnecessary, what could be better is to have a reset to default.

backend/game/values/constants.py
```diff
@@ -8,6 +8,7 @@
 DEATH_HIT_POINT = 0
 DEFAULT_ENERGY_CUBE = 0
 CARD_STORE_SIZE_LIMITER = 3
+ZERO_ENERGY = 0
```
gpfaia: I am starting to question some of these constants.   Default Energy i think would be more appropriate then zero.

backend/game/player/player.py
```diff
@@ -9,6 +9,7 @@ def __init__(self):
         self.is_alive = True
         self.victory_points = constants.DEATH_HIT_POINT
         self.energy = constants.DEFAULT_ENERGY_CUBE
+        self.card_on_hand = {}
```
gpfaia: As stated in a previous review.... due to duplicate cards, a dictionary is not a valid way to store them

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
```diff
@@ -0,0 +1,22 @@
...
+    def immediate_effect(self, player_that_bought_the_card, other_players):
+        card = ItHasAChild()
+        player_that_bought_the_card.add_card(card)
+
+    def special_effect(self, player_that_bought_the_card, other_players):
+        if player_that_bought_the_card.energy == constants.ZERO_ENERGY:
+            player_that_bought_the_card.discard_all_cards()
+            player_that_bought_the_card.lose_all_stars()
+            player_that_bought_the_card.update_health_by(10)
+
```
gpfaia: This function would also needs to move the player to outside of Tokyo.

Vlek:
This now seems like Keith has finished this in another pull request. Should this one now be closed?

cvennel:
closes #115

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
```diff
@@ -0,0 +1,21 @@
+from game.cards.card import Card
+from game.cards.keep_card import KeepCard
+from game.values import constants
+
+
+class ItHasAChild(KeepCard):
+    def __init__(self):
+        Card.__init__(self, "It Has a Child", 7,
```
cvennel: Should this be calling super?

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
```diff
@@ -0,0 +1,21 @@
...
+                      "If you reach 0[health] discard all your cards and lose all you [star]. Gain 10[health] and continue playing outside Tokyo",
+                      "")
+
+    def immediate_effect(self, player_that_bought_the_card, other_players):
+        card = ItHasAChild()
+        player_that_bought_the_card.add_card(card)
+
+    def special_effect(self, player_that_bought_the_card, other_players):
+        if player_that_bought_the_card.current_health == constants.DEATH_HIT_POINT:
```
cvennel: Do we need a constant DEATH_HIT_POINT if it's just a constant for 0? almost seems more confusing since it is very unlikely death would ever occur not at 0.

cvennel: I feel like the comparison is if health is <=0

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
```diff
@@ -0,0 +1,21 @@
...
+            player_that_bought_the_card.discard_all_cards()
+            player_that_bought_the_card.lose_all_stars()
+            player_that_bought_the_card.leave_tokyo()
+            player_that_bought_the_card.update_health_by(10)
```
cvennel: if the health is < 0 update_health_by wouldn't set health to 10

backend/game/cards/keep_cards/health_manipulation_cards/test_it_has_a_child.py
```diff
@@ -0,0 +1,49 @@
...
+from game.values.locations import Locations
+
+
+def test_it_has_a_child_player_add_card(player):
+    ItHasAChild().immediate_effect(player, None)
+    assert player.has_instance_of_card(ItHasAChild())
+
+
+def test_it_has_a_child_costs_7_energy():
+    assert ItHasAChild().cost == 7
```
cvennel: seems like an unnecessary test

backend/game/cards/keep_cards/health_manipulation_cards/test_it_has_a_child.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_it_has_a_child_discards_all_cards_when_health_is_0(player):
+    player.current_health = 0
+    ItHasAChild().immediate_effect(player, None)
```
cvennel: I'm not a fan of card.immediate_effect meaning buy the card. It obfuscates the point of the method

backend/game/cards/keep_cards/health_manipulation_cards/test_it_has_a_child.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_it_has_a_child_moves_outside_tokyo_when_played(player):
+    player.move_to_tokyo()
+    player.current_health = 0
+    ItHasAChild().immediate_effect(player, None)
+    ItHasAChild().special_effect(player, None)
+    assert player.location == Locations.OUTSIDE
+
+
+def test_it_has_a_child_no_affect_when_health_above_zero(player):
```
cvennel: Maybe add a test for if health is < 0

backend/game/player/player.py
```diff
@@ -37,6 +37,12 @@ def update_energy_by(self, change_integer):
         if self.energy < constants.DEFAULT_ENERGY_CUBE:
             self.energy = constants.DEFAULT_ENERGY_CUBE
 
+    def lose_all_stars(self):
+        self.victory_points = constants.DEATH_HIT_POINT
+
+    def discard_all_cards(self):
+        self.cards = []
```
cvennel: Cards shouldn't just disappear, they need to go back to the discard pile

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
gpfaia: yes

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
gpfaia: i agree with this point

backend/game/cards/keep_cards/health_manipulation_cards/test_it_has_a_child.py
gpfaia: Making changes so this is no longer needed

backend/game/player/player.py
gpfaia: I agree though not sure where this discard pile exists, I'll make a change to use the remove method and list comprehension but then I think all cards that need to be removed should use a discard method on the card

cvennel: There's a `deckhandler.discard()` method, but not exactly sure how to attach that yet

PR #73, opened 2019-12-05
Even Bigger keep card

### LINKED ISSUE #100: Even Bigger
Cost: 4
Type: Keep
Effect: +2[health] when you buy this card. Your maximum [health] is increased to 12 as long as you own this card
Footnote:

kaubster:
closes #100 

"Even Bigger" 

Cost: 4
Type: Keep
Effect: +2[health] when you buy this card. Your maximum [health] is increased to 12 as long as you own this card
Footnote:


backend/game/cards/keep_cards/victory_points_manipulation_cards/eater_of_the_dead.py
```diff
@@ -0,0 +1,11 @@
+from game.cards.keep_card import KeepCard
+from game.cards.card import Card
+from game.values import constants
+
+
+class EaterOfTheDead(KeepCard):
+    def __init__(self):
+        KeepCard.__init__(self, "Eater of the Dead", 4, "+ 3[Star] every time a monster's [Heart] goes to 0.")
```
gpfaia: can just call super here

backend/game/cards/keep_cards/victory_points_manipulation_cards/test_eater_of_the_dead.py
```diff
@@ -0,0 +1,9 @@
+from game.cards.keep_cards.victory_points_manipulation_cards.eater_of_the_dead import EaterOfTheDead
+
+def test_eater_of_the_dead_adds_3_victory_points(player, other_players):
+    EaterOfTheDead().immediate_effect(player, None)
+    EaterOfTheDead().special_effect(player, other_players)
+    assert player.victory_points == 3
+
+def test_eater_of_the_dead_costs_4_energy():
+    assert EaterOfTheDead().cost == 4
```
gpfaia: Add line break at the end

backend/game/player/player.py
```diff
@@ -9,6 +9,7 @@ def __init__(self):
         self.is_alive = True
         self.victory_points = constants.DEATH_HIT_POINT
         self.energy = constants.DEFAULT_ENERGY_CUBE
+        self.card_on_hand = {}
```
gpfaia: Discussed this in another review, cannot have dictionary store cards because can be duplicates, probably better to use a list

backend/game/player/player.py
```diff
@@ -32,3 +38,21 @@ def update_energy_by(self, change_integer):
         self.energy += change_integer
         if self.energy < constants.DEFAULT_ENERGY_CUBE:
             self.energy = constants.DEFAULT_ENERGY_CUBE
+
+    def add_card(self, card):
+        self.card_on_hand[card.name] = card.effect
```
gpfaia: Not sure what the purpose of storing the card by its name and string of the effect in a dictionary like this

backend/game/cards/keep_cards/victory_points_manipulation_cards/eater_of_the_dead.py
kaubster: I'd be happy to fix this, but this code came from Eater_Of_The_Dead branch. My apologize for the dup, I should probably removed it from this PR.

backend/game/cards/keep_cards/victory_points_manipulation_cards/test_eater_of_the_dead.py
kaubster: reformatted file via PyCharm

backend/game/player/player.py
kaubster: yup, sorry. I will update my PR given player.py changes made to branch It_Has_A_Child

backend/game/player/player.py
kaubster: see comments within backend/game/player/player.py above...

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
```diff
@@ -0,0 +1,20 @@
...
+from game.cards.keep_card import KeepCard
+
+
+class EvenBigger(KeepCard):
+    def __init__(self):
+        super(KeepCard, self).__init__("Even Bigger", 4,
+                                         "Your maximum [Heart] is increased by 2. Gain 2[Heart] when you get this card",
+                                         "")
+
+    def immediate_effect(self, player_that_bought_the_card, other_players):
```
gpfaia: This is a bit confusing of a function as you define a method in the parent class, then overload it, but use the parent's original method in the body of child?  I think this could be refactored

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
```diff
@@ -0,0 +1,20 @@
...
+        KeepCard.immediate_effect(self, player_that_bought_the_card, other_players)
+        EvenBigger().special_effect(player_that_bought_the_card, None)
+
+    def discard_effect(self, player_that_bought_the_card, other_players):
+        KeepCard.discard_effect(self, player_that_bought_the_card, other_players)
+        player_that_bought_the_card.update_max_health_by(-2)
+
+    def special_effect(self, player_that_bought_the_card, other_players):
+        player_that_bought_the_card.update_max_health_by(2)
```
gpfaia: I believe it should not just add to the max health but add to the regular health as well.

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
kaubster: was attempting to reduce code duplication, however your right. I was hiding too much. Coupling to self of sorts. Much harder to maintain.

Added player.add_card() and player.remove_card() card calls to immediate_effect and discard_effect methods respectively to fix.

Updated test_even_bigger.py

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
kaubster: see https://github.com/borisv13/kot/pull/73#discussion_r355237931

backend/game/player/player.py
kaubster: you mentioned this was discussed, do you recall which PR?

Perhaps it was fixed already.

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class ItHasAChild(KeepCard):
+    def __init__(self):
+        super(KeepCard, self).__init__("It Has a Child", 7,
+                      "If you reach 0[health] discard all your cards and lose all you [star]. Gain 10[health] and continue playing outside Tokyo",
+                      "")
+
+    def special_effect(self, player_that_bought_the_card, other_players):
+        if player_that_bought_the_card.energy == constants.ZERO_ENERGY:
```
cvennel: I think this should be `player_that_bought_the_card.current_health` not `player_that_bought_the_card.energy`

backend/game/cards/keep_cards/health_manipulation_cards/test_even_bigger.py
```diff
@@ -0,0 +1,38 @@
...
+
+def test_even_bigger_gains_2_max_health_when_purchased(player):
+    EvenBigger().immediate_effect(player, None)
+    assert player.maximum_health == constants.DEFAULT_HEALTH + 2
+
+def test_even_bigger_gains_2_current_health_when_purchased(player):
+    EvenBigger().immediate_effect(player, None)
+    assert player.current_health == constants.DEFAULT_HEALTH + 2
+
+def test_it_has_a_child_costs_4_energy():
```
cvennel: test name points to wrong card

backend/game/player/test_player.py
```diff
@@ -93,3 +97,13 @@ def test_player_energy_cannot_go_below_zero(player):
...
+
+
+def test_player_health_cannot_exceed_decreased_max_health(player):
+    player.update_max_health_by(-1)
+    assert player.current_health == player.maximum_health
+
+
+def test_player_health_less_than_increased_max_health(player):
+    player.update_max_health_by(1)
+    assert player.current_health < player.maximum_health
```
cvennel: Should we add tests for the player keep card methods?

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class EvenBigger(KeepCard):
+    def __init__(self):
+        super(KeepCard, self).__init__("Even Bigger", 4,
+                                         "Your maximum [Heart] is increased by 2. Gain 2[Heart] when you get this card",
+                                         "")
+
+    def immediate_effect(self, player_that_bought_the_card, other_players):
+        player_that_bought_the_card.add_card(self)
```
Vlek: I believe this is something that the professor would rather be handled at the game level rather than having the card manage the player's inventory.

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
Vlek: @cvennel is right. Based on the description, this should take affect when they player's health reaches zero.

backend/game/player/player.py
Vlek: Having a class for a hand seems silly and doesn't really address the issue if it's going to be a dictionary. It likely should be a list. The only issue may be that, if the player has two and wanted to discard the second one, the first may disappear which can be unsettling. We would then have to have something like a manual option which takes a card index.

This also gives me an interesting thought. Are there any cards that, if they have a special affect that happens, could cause another card to use its special affect? So like, if a player has a card they use, should it then go through their cards to ensure that there are no other special affects that should happen? It may also require that everyone's cards are checked for special affects if they can play off each other like that.

cvennel:
closes #100

backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
kaubster: moved "it has a child" to its own PR

backend/game/cards/keep_cards/health_manipulation_cards/test_even_bigger.py
kaubster: fixed test name typo

backend/game/player/test_player.py
kaubster: added tests

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
kaubster: removed add/remove card from player logic... agree was bad coupling

kaubster:
closes #100

backend/game/cards/keep_card.py
```diff
@@ -7,11 +7,11 @@ def __init__(self, name, cost, effect):
         super().__init__(name, cost, effect)
 
     def immediate_effect(self, player_that_bought_the_card, other_players):
-        player_that_bought_the_card.add_card(self)
+        pass
```
kaubster: removed internal adding and removing of card to player... thinking higher order centralized place should add card to players hand... but could be wrong.

Can put back if you all prefer.

backend/game/cards/keep_card.py
```diff
@@ -7,11 +7,11 @@ def __init__(self, name, cost, effect):
...
 
     @abstractmethod
     def special_effect(self, player_that_bought_the_card, other_players):
         raise NotImplementedError
 
     def discard_effect(self, player_that_bought_the_card, other_players):
-        player_that_bought_the_card.remove_card(self)
\ No newline at end of file
+        pass
```
kaubster: removed internal adding and removing of card to player... thinking higher order centralized place should add card to players hand... but could be wrong.

Can put back if you all prefer.

backend/game/cards/keep_card.py
cvennel: Gerardo and I agree with these being removed, however are planning to handle this in a future refactor

backend/game/cards/keep_card.py
cvennel: Gerardo and I agree with these being removed, however are planning to handle this in a future refactor

backend/game/cards/keep_cards/health_manipulation_cards/even_bigger.py
```diff
@@ -0,0 +1,19 @@
+from game.cards.card import Card
+from game.cards.keep_card import KeepCard
+
+
+class EvenBigger(KeepCard):
+    def __init__(self):
+        super(KeepCard, self).__init__("Even Bigger", 4,
```
gpfaia: 1. Not the proper way to call super
2. Argument count is incorrect

backend/game/cards/keep_cards/health_manipulation_cards/test_even_bigger.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+
+from game.cards.keep_cards.health_manipulation_cards.even_bigger import EvenBigger
+from game.player.player import Player
+from game.values import constants
+
+
+@pytest.fixture(autouse=True)
```
gpfaia: you do not need to add this fixture/def here

backend/game/cards/card.py
```diff
@@ -7,3 +7,7 @@ def __init__(self, name, cost, effect, footnote=None):
         self.cost = cost
         self.effect = effect
         self.footnote = footnote
+
+    # Card is a user defined object. Provided comparison context other than object reference.
```
gpfaia: question: is this used as a part of this changelist?

backend/game/cards/keep_cards/health_manipulation_cards/test_even_bigger.py
cvennel: I did this so I didn't have to keep doing the add player, add the card to the player, then trigger the immediate effect. Should I just have done that in each test?

backend/game/deck/deck_handler.py
```diff
@@ -68,6 +68,7 @@ def buy_card_from_store(self, index, purchasing_player: Player):
             purchasing_player.add_card(card_to_buy)
             self.__card_store.remove(card_to_buy)
             self.__fill_card_store()
+        return card_to_buy
```
gpfaia: Not sure if this is expected or not

backend/game/cards/card.py
cvennel: Player.remove card needs this to function

backend/game/deck/deck_handler.py
cvennel: Something keith added, I didn't have a problem with it so I left it. Might be to attach to front end.

PR #74, opened 2019-12-08
added a register user form on the login page. it redirects to the lob…

CameronBaffuto:
…by at the moment but will be connected when user authentication and account creation is ready

### DIFF
--- a/frontend/src/containers/RegisterUser.js
+++ b/frontend/src/containers/RegisterUser.js
@@ -1,0 +1,75 @@
+import React, { Component } from 'react';
+import { Redirect } from 'react-router-dom'
+import Button from 'react-bootstrap/Button'
+import "./Login.css";
+import Form from "react-bootstrap/Form";
+
+class RegisterUser extends Component {
+  constructor(props){
+    super(props);
+    this.state={
+      first_name:'',
+      last_name:'',
+      email:'',
+      password:''
+    }
+  }
+
+  setRedirect = () => {
+        this.setState({
+            redirect: true
+        })
+    }
+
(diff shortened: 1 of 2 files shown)


frontend/src/containers/RegisterUser.js
```diff
@@ -0,0 +1,75 @@
...
+             type="email"
+             floatingLabelText="Email"
+             onChange = {(event,newValue) => this.setState({email:newValue})}
+             />
+             <Form.Label>Password</Form.Label>
+           <Form.Control
+             type = "password"
+             placeholder="Enter your Password"
+             floatingLabelText="Password"
+             onChange = {(event,newValue) => this.setState({password:newValue})}
```
kaubster: Suggest adding a method instead of using an anomalous method here. Method will be important later when we need hash the password prior to storing in state.

frontend/src/containers/RegisterUser.js
```diff
@@ -0,0 +1,75 @@
...
+    this.state={
+      first_name:'',
+      last_name:'',
+      email:'',
+      password:''
+    }
+  }
+
+  setRedirect = () => {
+        this.setState({
```
kaubster: Small comment for future posterity.

Suggest adding a comment within method setRedirect() just prior to this.setState...

/* Add async exchange to determine login availability and eventually success/failure. Suggest using Auth0... example: https://auth0.com/blog/handling-authentication-in-react-with-context-and-hooks/ */

In the future we will not want method renderRedirect() to redirect unless the server has provided an auth token...

PR #75, opened 2019-12-08
Db pair work

otaiwo70:


cvennel:
Investigating broken build

cvennel:
@otaiwo70 The build is failing due to pytest test discovery. Pytest works from backend but not from the root directory. I've been trying to fight this with no luck.

@borisv13 it looks like pytest is only being used for backend tests, is that accurate? If so can we change the build script to point at the backend?  `py.test  --cov-report term --cov=.` fails but I think this would work with `py.test  backend --cov-report term --cov=.`  (it at least solves the problem on my machine) This would also greatly speed up the test process for CI. 

The alternative I was playing with was having multiple settings files or pytest.ini files but then I was running into issues with the root perspective changing which messed up file imports since they're using relative paths

cvennel:
@otaiwo70 You also may want to mark this PR as a WIP and tag others to weigh in.

cvennel:
Updated Travis to xenial to allow later sqlite3 version, added script steps to change PYTHONPATH to backend, run pytest from backend, and setup django... Still no luck. I'm giving up for the evening.

cvennel:
@otaiwo70 Is this all outdated? If so can we close this one?

PR #77, opened 2020-01-16
Message send with server response

kaubster:
Kanban (1of2):  Message Send with server Response (dice and console messages)
         https://github.com/borisv13/kot/projects/1#card-27893374
Task 1 - placeholder created and placed on Gameboard UI, basic tests added.
Task 2 - Integrate/display game engine responses from backend

Kanban (2of2): Send with server Response (send dice and console messages)
         https://github.com/borisv13/kot/projects/1#card-31437344 

Acceptance Criteria:
- frontend send message selected_dice with payload content proving list of dice
- backend receives message
- backend echos "server_response" back to front end 
- frontend display payload within GameLog
- Add/Update User Record in DB
- Add Player after initial connection with game
- When existing, retrieve existing player

Commands added...
         'init_chat_request' - initialize backend player and initial game room state
   Game Console commands:
         'gamelog_send_request' - send command using console "submit" button or hitting [ENTER] to backend
         'selected_dice_request' - send all dice and which were selected to backend


.vscode/launch.json
```diff
@@ -1,77 +1,58 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-    "version": "0.2.0",
-    "configurations": [
+  // Use IntelliSense to learn about possible attributes.
```
Vlek: Please remove these files and gitignore them.

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "python.pythonPath": "backend\\env\\Scripts\\python.exe",
+    "python.pythonPath": "env\\Scripts\\python.exe",
```
Vlek: This entire folder should be ignored.

README.md
```diff
@@ -10,79 +10,56 @@ Frontend documentation see \kot\frontend\README.md.
...
+- Development: run React ('npm start') and Python ("python manage.py runserver") seperately. Allows for React and python to auto watch and reload.
 
 ## Site Map
 
-* <http://localhost:8000/>
-* <http://localhost:8000/api/v1/dice/>
-* <http://localhost:8000/admin/>
-* <http://localhost:8000/users/>
+- <http://localhost:8000/>
+- <http://localhost:8000/api/v1/dice/>
```
Vlek: This should really be removed. This isn't how the game will use dice and was something that you added as a separate thing to test sending and receiving messages on the front end.

frontend/src/components/Cards/CardStore.js
```diff
@@ -1,106 +1,110 @@
...
-            cardFootnote: 'footnote description',
-
-        };
-    }
+  constructor(props, context) {
+    super(props, context);
+
+    this.handleChange = this.handleChange.bind(this);
+    this.clearSelected = this.clearSelected.bind(this);
+
```
Vlek: This is now no longer about the backend stuff and should likely have been separated into a separate pull request.

.vscode/launch.json
kaubster: "Keith has done a lot and taken several steps without really checking in and explaining that was the way that he was going to do things. " - this was marked as a draft for the purpose of discussion, so honestly thank you for the for the feedback!

"creating an entire new class to have as a message instead of attempting to deal with the information as it comes in"
- Not sure this is more complex, other than it actually catches and parses messages by...
- Single entry-point for all messages seemed like the most efficient approach
- commands are defined in a single commands array. No need for switch statements to add new commands. Add the command and a handler.
- Given single responsibility, each handler should be in a separate class within a final PR.

.vscode/settings.json
kaubster: sure, final PR could exclude these files. They were in master so I inherited them.

README.md
kaubster: Rolling dice is a stateless operation, so one off endpoints calls seems fine. Once we have a command for sending dice (which would be a dup of effort) we can disable this I guess.

frontend/src/components/Cards/CardStore.js
kaubster: yep, inherited. Defn would not be in final PR ... unless its in master.

.travis.yml
```diff
@@ -11,7 +11,13 @@ python:
 install:
   - pip install -r requirements.txt
 script:
-  - py.test  --cov-report term --cov=.
+  - echo "TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST"
```
kaubster: totally playing with travis here.. this will not be in a final PR

kaubster:
will be deleting this PR soon.... broken out in to PRs #151 and #152

PR #78, opened 2020-01-20
NovaBreath Keep card 

cvennel:
Copied the Keep card abstract class I think from @kaubster PR #73.

Since this card only modifies the targets of attack rolls no trigger logic is necessary, just check if attacker owns this card

### DIFF
--- a/backend/game/cards/keep_card.py
+++ b/backend/game/cards/keep_card.py
@@ -1,0 +1,17 @@
+from abc import abstractmethod
+from game.cards.card import Card
+
+
+class KeepCard(Card):
+    def __init__(self, name, cost, effect):
+        super().__init__(name, cost, effect)
+
+    def immediate_effect(self, player_that_bought_the_card, other_players):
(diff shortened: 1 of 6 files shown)


cvennel:
#82

cvennel:
#123

cvennel:
closes #123

backend/game/cards/keep_card.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def immediate_effect(self, player_that_bought_the_card, other_players):
+        player_that_bought_the_card.add_card(self)
+
+    @abstractmethod
+    def special_effect(self, player_that_bought_the_card, other_players):
+        raise NotImplementedError
+
+    def discard_effect(self, player_that_bought_the_card, other_players):
+        player_that_bought_the_card.remove_card(self)
```
gpfaia: NIT: Best practice is to have a new line at EOF

PR #79, opened 2020-01-20
Pytest test cases for the model

otaiwo70:
Below are the pytest test cases for the model db

### DIFF
--- a/backend/game/modelPlayTest.py
+++ b/backend/game/modelPlayTest.py
@@ -1,0 +1,58 @@
+
+import datetime
+
+from django.test import TestCase
+
+from game.models import Dice
+from game.models import Game
+from game.models import Play
+from game.models import User
+
+
+class PlayModelTest(TestCase):
+
+    @classmethod
+    def setUp(cls):
+        user = User.objects.create(monster_name='Godzilla',
+                                   username='User1',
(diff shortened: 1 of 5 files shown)


backend/game/modelPlayerTest.py
```diff
@@ -0,0 +1,20 @@
...
+from game.models import User
+
+
+class PlayerModelTest(TestCase):
+
+    @classmethod
+    def setUp(cls):
+        User.objects.create(monster_name='Godzilla',
+                            username='User1',
+                            password='Password1',
```
kaubster: we removing passwords?

backend/game/modelPlayTest.py
```diff
@@ -0,0 +1,58 @@
...
+                                   date_created=datetime.datetime.now())
+
+        game = Game.objects.create(user=user,
+                                   is_winner='Y',
+                                   num_players='10',
+                                   player_position='1',
+                                   date_created=datetime.datetime.now(),
+                                   date_modified=datetime.datetime.now())
+
+        dice = Dice.objects.create(user=user,
```
kaubster: there are many hard-coded repeating values in this class... are these expected data table specific values? If so, would it make sense to add constants for each?

PR #80, opened 2020-01-20
Db crud

otaiwo70:
first draft for ongoing work on player using repository pattern


backend/game/player/test_player.py
```diff
@@ -13,6 +13,21 @@ def player():
...
+    assert player.monster_name == 'Godzilla'
+
+
+def test_set_username(player):
+    player.set_username('Tester')
+    assert player.username == 'Tester'
+
+
+def test_set_password(player):
+    player.set_password('12345678')
```
cvennel: I don't think we should be storing plain text passwords for users. Are we planning to implement authentication? If so l, I think we should handle that outside of the player class. I don't think passwords should go in these database objects and can just be removed for the scope of this PR. If we used some external authentication provider (like sign in with google) we wouldn't even need passwords I would think

otaiwo70: password reference removed

backend/game/migrations/0001_initial.py
```diff
@@ -59,7 +61,8 @@ class Migration(migrations.Migration):
                 ('monster_name', models.CharField(max_length=30)),
                 ('username', models.CharField(max_length=30)),
                 ('password', models.CharField(max_length=30)),
```
kaubster: remove passwords?

otaiwo70: True removed previously

otaiwo70: Migration was not done but redid the migration to reference updated model my mistake

PR #147, opened 2020-01-23
engine/board.py tweaks

cvennel:
- Added tests to game board
- fixed unresolved reference error in board.py where players should have been self.player
- changed add_player to not be a protected method in Player_queue.py


backend/game/engine/player_queue.py
```diff
@@ -10,7 +10,7 @@ def __init__(self):
         self.current_player = None
         self.player_cycle = []
 
-    def _add_player_to_game(self, new_player):
```
cvennel: @gpfaia I'm not seeing why this was a protected method. If this should be protected let me know and we can change it back

gpfaia: No originally just wanted people to be mindful of adding players after the game was started was the original intention

backend/game/engine/test_board.py
```diff
@@ -12,3 +13,53 @@ def test_after_game_starts():
...
+def test_get_next_player_turn_consistent_order():
+    game = BoardGame()
+    for _ in range(3):
+        game.add_player(Player())
+    game.start_game()
+    player1 = game.get_next_player_turn()
+    player2 = game.get_next_player_turn()
+    player3 = game.get_next_player_turn()
+
+    for _ in range(50):
```
gpfaia: Probably a bit excessive but gets the point across

cvennel: Yeah, was definitely excessive

cvennel:
@gpfaia  just had to resolve merge conflicts which invalidated your approval. While I was at it, changed that excessive 50 times loop to 5

PR #148, opened 2020-01-23
demonstrate attach deckhandler to frontend GameConsumer

cvennel:
Sample of attaching deckhandler to gameconsumer.
NOTE: this PR doesn't merge to Master, but in to Lakshmi's branch, I just did this to separate so it can be applied as needed. 

Keith I just added you since you were on the call in case you're interested

### DIFF
--- a/backend/lobby/consumers.py
+++ b/backend/lobby/consumers.py
@@ -1,8 +1,13 @@
-from channels.generic.websocket import WebsocketConsumer
 import json
+
+from channels.generic.websocket import WebsocketConsumer
+
+from backend.game.engine.board import BoardGame
 
 
 class GameConsumer(WebsocketConsumer):
+    master_game_controller = BoardGame()
+
     def connect(self):
         self.accept()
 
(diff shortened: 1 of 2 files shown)


backend/lobby/consumers.py
```diff
@@ -25,7 +30,23 @@ def randomcards_deck_handler(self, data):
         cards_required = data['payload']
         self.send_server_response_to_client('GAMECONSOLE_DECISION', username, room, cards)
 
+    def get_the_three_store_cards(self):
```
kaubster: Is it possible to add a parameter so Lakshmi can request 0 or more cards?

Her message includes a cards attribute where she specifies a number of requested cards.

kaubster:
Other than the little tweak, looks good.

backend/lobby/consumers.py
cvennel: I just wrote that to get the whole card store then in her handling of the result I figured she could filter what she wanted. I think getting the full card store each time would probably be easier because if you buy the middle or first card then the other slide over and the deck handler automatically draws a third card. If the draw pile runs out the deck handler automatically shuffles the discard pile etc. So just copying the full state of the card store means less logic on the frontend

cvennel: @kaubster not sure if my reply automatically pings you

kaubster: Ok, I like that approach. Sold....

backend/lobby/consumers.py
```diff
@@ -1,8 +1,13 @@
-from channels.generic.websocket import WebsocketConsumer
 import json
 
+from channels.generic.websocket import WebsocketConsumer
+
+from backend.game.engine.board import BoardGame
+
 
 class GameConsumer(WebsocketConsumer):
+    master_game_controller = BoardGame()
```
kaubster: The channels changes made within branch 'add_channel_roll_cmd' will have you accessing the latest BoardGame instance for the room via DB...

We do not have a PR for this yet. Once we do or its merged I can let you know.

backend/lobby/consumers.py
```diff
@@ -25,7 +30,23 @@ def randomcards_deck_handler(self, data):
...
+
     commands = {
         'randomcards_send_request': randomcards_deck_handler
-        #To modify the command with backend class that handles this
-    }
\ No newline at end of file
+    }
+
+
+# This test is temporary to prove the GameConsumer has access to the card store
```
kaubster: can this be moved to a new file "consumer_test.py"?

PR #149, opened 2020-01-23
Refactor card store to generate cards dynamically

Lakshmii3:
1. CardStore.js reuses <Card> element repeatedly. Refactored code such that cards will be generated dynamically from an array of selected cards received from backend.
2. Added test cases and updated snapshots.


frontend/src/components/Cards/CardStore.js
```diff
@@ -5,27 +5,26 @@ import Button from 'react-bootstrap/Button';
...
+    this.handleChange = this.handleChange.bind(this);
+    this.clearSelected = this.clearSelected.bind(this);
+
+    this.state = {
+      username: props.currentUser,
+      gameRoom: props.currentRoom,
+      value: [],
+      selectedCard: [],
+      card: [
+        { name: 'Armor Plating', cost: 4, type: 'Keep', effect: 'Ignore damage of 1', footnote: ' ' },
```
cvennel: I don't understand why these are here. Don't the cards get generated now? Wouldn't that mean we don't need these hard coded values?

Lakshmii3: I added those hard coded values before changes were made to engine/board.py. Removed those values

PR #153, opened 2020-01-26
removed discard and add methods from keepcard

cvennel:


cvennel:
I feel like these methods shouldn't exist in the keep card methods. The player class can track their own cards

gpfaia:
> 
> 
> I feel like these methods shouldn't exist in the keep card methods. The player class can track their own cards

We discussed this issue in the chat and at first I did agree but after working on a card that had all cards discarded, the discard effect was helpful.  Either way I think we will need something that takes discard cards from the player to a discard list so although I think this could be good, unsure of what the right implementation is right now.  The immediate effect I probably agree with getting rid of though

cvennel:
> > I feel like these methods shouldn't exist in the keep card methods. The player class can track their own cards
> 
> We discussed this issue in the chat and at first I did agree but after working on a card that had all cards discarded, the discard effect was helpful. Either way I think we will need something that takes discard cards from the player to a discard list so although I think this could be good, unsure of what the right implementation is right now. The immediate effect I probably agree with getting rid of though

I believe the deck handler has a discard card method with an optional argument for where it comes from and removes from that list, could that be what you're looking for?

cvennel:
I think this should just close and we start fresh with the refactor

PR #154, opened 2020-01-26
repository pattern for message model

otaiwo70:
repository pattern for message model


otaiwo70:
closing because message model was deleted by #151

PR #160, opened 2020-02-01
repository pattern to support play

otaiwo70:
repository pattern to support play


backend/game/irepository/irepository_player.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class IRepositoryPlayer:
+    def __init__(self):
+        self.user = None
+
+    def save_player_db(self, player: Player):
+        self.user = User(monster_name=player.monster_name,
+                         username=player.username,
+                         password=player.password,
```
cvennel: Don't think passwords should be save plaintext see review on #80

otaiwo70: rectified by using password=md5.encrypt(player.password),

otaiwo70: password reference removed

kaubster: will have to deal with passwords eventually... but agree with removing for the moment.

PR #163, opened 2020-02-02
Mt edits for frontend connection

cvennel:


backend/game/player/player.py
```diff
@@ -6,14 +7,21 @@
 
 
 class Player:
-    def __init__(self):
+    def __init__(self, username=None):
```
gpfaia: Something you can do here is set the default to "guest_{}".format(random.randint(1000, 9999)) that way you can avoid the if statement, since you know what the default is and its never going to be None

backend/game/engine/dice_msg_translator.py
```diff
@@ -0,0 +1,22 @@
...
+            indexes.append(i)
+        i += 1
+    return indexes
+
+
+def dice_values_message_create(values):
+    msg = []
+    for val in values:
+        translated_val = translate_enum_to_frontend_val(val.name)
+        msg.append([translated_val, False])
```
gpfaia: Why is this being sent with false as the 2nd index?

cvennel: This is a front-end compatibility thing. I should add that to a description. Basically it sets front-end to be "not selected" for the re-roll UI.

backend/game/player/player.py
cvennel: That's what I tried at first. The problem is that random number is generated one time only so all players had the same random number for every test. The random number generation had to be moved inside the class so it would wait until object creation to actually get a random number. 

my first attempt was `def __init__(self, username="guest_{}".format(random.randint(1000, 9999))` but all the tests broke

backend/game/engine/dice_msg_translator.py
cvennel: Added method explanation

cvennel:
This PR needs updating with the bugfixes from last night before merging

PR #170, opened 2020-02-08
Quick bugfix

cvennel:
I noticed a small mistake in dice resolver. We should be moving to tokyo if empty on rolling at least one die, not > 1 die.


cvennel:
closes #171

Issue #145, opened 2020-01-21
We're Only Making It Stronger


cvennel:
Cost: 3
Type: Keep
Effect: When you lose 2[health] or more gain 1[energy]
Footnote:

PR #176, opened 2020-02-09
We're only makng it stronger

gpfaia:
Added were only making it stronger card/tests
Updated player health model to account for card

 #145


gpfaia:
resolves #145, merge conflicts fixed

PR #185, opened 2020-02-13
Display player cards

kaubster:
NOTE: To avoid reviewing generated files for testing, Filter on .js, .css and .py prior to starting review.

Now display cards next to player stats
local player stats appear to right of game console
opponent stats appear along bottom of board
No cards results in an "No Cards message" next to card icon.
Cards display energy and popup tooltips displaying effect and footnote

I did move Discard.type and KeepCard.type up to Card.Type. It was nice to have type on the frontend. But type was not guaranteed useful/set unless it becomes required by Card. I'm open to other suggestions if you all do not agree of course.


cvennel:
I can't get  the server  to run on this branch :/ I'm not sure what's going on

cvennel:
A previous merge broke the backend/frontend interaction. Fixed it, merged in master

PR #186, opened 2020-02-14
commenting out irepository pytest

otaiwo70:
commenting out irepository pytest


cvennel:
Are we not using this any more? Why not just remove? What exactly is this for?

PR #192, opened 2020-02-19
Sweep store feature integration with backend

Lakshmii3:
1) Added sweep store button
2) Player can sweep store if he has sufficient energy
3) Player cannot sweep store if there is not enough energy

ToDo:
To only enable the Sweep Store button for player who has the current turn to play


cvennel:
Don't merge yet.... I broke this

PR #193, opened 2020-02-21
store of game to db

otaiwo70:
update to consumers.py is possibly how i see this working refer to section below

        if not IRepositoryGame.get_game_by_room(room):
            game_store = IRepositoryGame.save_game(room)
            print("Room created with id" + game_store)
        else:
            error = 'Room already exist, Unable to create Game with room: ' + room


backend/game/irepository/test_irepository_dice.py
```diff
@@ -1,86 +1,86 @@
...
-    repository_game.save_game('Room5')
-    repository_player = IRepositoryPlayer()
-    player = Player()
-    player.set_monster_name('Godzilla1')
-    player.set_username('Test_Get')
-    repository_player.save_player(player, 'Room5')
-    repository_dice.save_dice(player, 'Room5', DieValue.ONE, DieValue.TWO, DieValue.THREE, DieValue.ONE,
-                              DieValue.TWO, DieValue.THREE, 0, 0, 1)
-    assert repository_dice.delete_dice_by_id(5) is None
+# import pytest
```
cvennel: Is this all just commented out to allow pytest to work on our local machines? If so I don't think you need to do that. I would argue to leave the code in, and let people know they should run ` py.test backend --ignore backend/game/db --ignore backend/game/irepository` so this doesn't look like  unnecessary code. Others may disagree with me though

kaubster: Agree, though it kind if stinks we need remember to did this.

kaubster: Agree, though it kind if stinks we need remember to did this.

otaiwo70: true did that because keith requested earlier, will add back as you both agree to do so

gpfaia: I added a line if you are using VS code that will automatically remove these tests; however, if using PyCharm or another software unsure if it will be excluded automatically

otaiwo70: Great and works for me, will merge once approved

Will need one approval to move on to merge

Issue #166, opened 2020-02-06
Improve game messages


cvennel:
Currently the latest message is on the bottom. As the game progresses players will have to scroll back and forth between the UI buttons at the top and the messages at the bottom to keep track of the game. 

Suggest at least flipping the messages to have newest message on top.

kaubster:
attempted to fix this via PR #196... added height limit so scrollbar appears.

PR #196, opened 2020-02-22
added lobby list of games

kaubster:
Added list of games to lobby

Each row in the list includes room name and players.


backend/lobby/consumers_lobby_test.py
```diff
@@ -0,0 +1,108 @@
+# import pytest
```
gpfaia: Is there a reason for a completely commented out test file?

kaubster:
added fix for issue #166

backend/lobby/consumers_lobby_test.py
kaubster: these tests work locally, but for some reason (I've not determined) these Django tests choke on the server.

Guess I can remove them though.

cvennel: I don't think tests that fail on the server should just be completely deleted, they still have value run locally. Instead of deleting you can just add an ignore argument to the Travis.yml py.test command. This way we still get the benefit of having the tests available but don't bother with the Django tests on Travis

Issue #201, opened 2020-02-23
Suspect end turn button now skips resolving dice. 


cvennel:
End turn should be blocked until dice resolved or preferably should also resolve dice

PR #203, opened 2020-02-24
Update end turn button to call accept dice if not already accepted

cvennel:


cvennel:
fixes #201

cvennel:
should get fixed by #204 after merging.

PR #215, opened 2020-02-29
fixed players hand UI update after purchasing card

### LINKED ISSUE #214: Player Values Hand does not show card after purchase
Purchased keep cards should appear in the players (who buys) hand.

Currently the UI does not update though data is sent from backend.

kaubster:
Fixes issue #214, hand now updates after purchasing keep card

closes #214

When reviewing... look at *.js files... there are only 3


gpfaia:
As a note, will this coverage commits mess up other users who do not have your pathing.  I don't believe so but just wanted to ask

PR #217, opened 2020-02-29
lock buy cards to current player

### LINKED ISSUE #216: Opposing player can force current player to buy cards
uh oh.

cvennel:
fixes #216


backend/lobby/consumers.py
```diff
@@ -217,6 +217,10 @@ def resolve_dice_handler(self, data):
     def buy_card_request_handler(self, data):
         username, room, game, state = reconstruct_game(data)
         index_to_buy = data['payload']
```
gpfaia: NIT: You can technically remove this extra spacing and then move the index_to_buy to after the bail out return.

cvennel: I kind of like the space to separate that thought from the rest, but do like the suggestion of moving the index

PR #219, opened 2020-02-29
Yield improvements

cvennel:
So... I totally forgot to add commit messages while working on this, so that's why there's two with an old commit message from another PR.  oops.


cvennel:
ezgif com-video-to-gif

cvennel:
closes #197

PR #220, opened 2020-02-29
Logging improvements

cvennel:
Sends log messages to the game chat window regarding rolls and card purchases so players can follow game progress


backend/lobby/consumers.py
```diff
@@ -12,11 +12,20 @@
...
 from lobby.consumers_common import save_game, reconstruct_game, create_send_response_to_client
 from lobby.server_message_types import PLAYER_STATUS_UPDATE_RESPONSE, BEGIN_TURN_RESPONSE, SERVER_RESPONSE, \
     DICE_ROLLS_RESPONSE, CARD_STORE_RESPONSE
 
 
+def dice_vals_log_message(player_name, values):
+    msg = "{} rolled: ".format(player_name)
+    for val in values[:-1]:
+        msg += val.name + ", "
+    msg += values[-1].name
```
gpfaia: This can be rewritten to be simpler/1 line

`msg = "%s rolled: %s" % (player_name, ", ".join([val.name for val in values]))`

you might have to wrap the val.name in str() though

PR #221, opened 2020-02-29
Frenzy discard card

### LINKED ISSUE #105: Frenzy
Cost: 7
Type: Discard
Effect: Take another turn after this one
Footnote:

cvennel:
closes #105


backend/game/cards/discard_cards/turn_manipulation_cards/frenzy.py
```diff
@@ -0,0 +1,10 @@
+from game.cards.discard_card import DiscardCard
+
+
+class Frenzy(DiscardCard):
+    def __init__(self):
+        super().__init__("Frenzy", 7,
+                         "Take another turn after this one", "")
```
gpfaia: I had previously not been putting anything for footnote field if it was empty

backend/game/cards/master_card_list.py
```diff
@@ -14,6 +14,7 @@
 from game.cards.discard_cards.multi_manipulation_cards.vast_storm import VastStorm
 
 # Victory Point Manipulation Cards
+from game.cards.discard_cards.turn_manipulation_cards.frenzy import Frenzy
```
gpfaia: It's fine to put it in here but the comment makes this misleading

cvennel: Ha totally missed that thanks

backend/game/cards/discard_cards/turn_manipulation_cards/frenzy.py
cvennel: Good call.

backend/game/cards/master_card_list.py
cvennel: auto import got me

cvennel: Resolved, but github isn't showing the outdated tag as the change is sorta in a comment

backend/game/engine/player_queue.py
```diff
@@ -36,7 +36,10 @@ def reset_allowed_to_yield(self):
 
     def get_next_player(self):
         self.reset_allowed_to_yield()
-        self.current_player = self._cycle_next_alive_player()
+        if self.current_player is not None and self.current_player.gets_bonus_turn:
```
gpfaia: this can be reduced. Unless can think of a specific reason for it "is not None" is often not needed as we can just check for it using a regular if

`if self.current_player and self.current_player.gets_bonus_turn:`

cvennel: Yeah that makes sense. Thanks

PR #223, opened 2020-03-01
Adding trigger for 'it has a child' on death

### LINKED ISSUE #157: Add trigger on death for 'Eater of Dead'
This card needs to know anytime a player's health reaches 0 to add victory points

gpfaia:
Add to master card list
Fix tests
closes #157

### DIFF
--- a/backend/game/cards/discard_cards/energy_manipulation_cards/energize.py
+++ b/backend/game/cards/discard_cards/energy_manipulation_cards/energize.py
@@ -3,7 +3,7 @@ from game.cards.discard_card import DiscardCard
 
 class Energize(DiscardCard):
     def __init__(self):
-        super().__init__("Drop from High Altitude", 8, "+ 9[Energy]", "")
+        super().__init__("Energize", 8, "+ 9[Energy]")
 
     def immediate_effect(self, player_that_bought_the_card, other_players):
         player_that_bought_the_card.update_energy_by(9)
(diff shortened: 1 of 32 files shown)


backend/game/cards/keep_cards/health_manipulation_cards/it_has_a_child.py
```diff
@@ -7,17 +7,9 @@ def __init__(self):
...
                          "If you reach 0[health] discard all your cards and lose all you [star]. Gain 10[health] and continue playing outside Tokyo.")
 
-    def immediate_effect(self, player_that_bought_the_card, other_players):
-        card = ItHasAChild()
-        player_that_bought_the_card.add_card(card)
-
     def special_effect(self, player_that_bought_the_card, other_players):
         if player_that_bought_the_card.current_health <= 0:
-            self._discard_all_cards(player_that_bought_the_card, other_players)
+            player_that_bought_the_card.discard_all_cards()
```
cvennel: The discard all cards method just clears the list effectively destroying all the cards. The cards should be discarded back to the deckhandler instead

gpfaia: Yes realizing this is going to need to be marked as a larger bug as this is how the discard method works for ALL cards currently (both keep and discard)

gpfaia: This is marked in #226 and can be handled outside of this change list as it affects much more than just this one card

Issue #224, opened 2020-03-01
Drop from high altitude is incorrect


cvennel:
Running through tests of the game I keep seeing "Drop from High Altitude" gives 9 energy. That's actually what Energize does. These two cards need to be investigated for accuracy

PR #225, opened 2020-03-01
Various card fixes

gpfaia:
Added keep card does proper super
Added NovaCard to own folder type (fixed its references)
Added back drop from high altitude
Fixed energize card name

### DIFF
--- a/backend/game/cards/discard_cards/victory_point_manipulation_cards/drop_from_high_altitude.py
+++ b/backend/game/cards/discard_cards/victory_point_manipulation_cards/drop_from_high_altitude.py
@@ -1,0 +1,16 @@
+from game.cards.discard_card import DiscardCard
+from game.values.locations import Locations
+import game.turn_actions.player_movement as player_movement
+
+
+class DropFromHighAltitude(DiscardCard):
+    def __init__(self):
+        super().__init__("Drop from High Altitude", 5,
+                         "+ 2[Star] and take control of Tokyo if you don't already control it")
+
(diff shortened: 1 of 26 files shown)


backend/game/cards/discard_cards/victory_point_manipulation_cards/drop_from_high_altitude.py
```diff
@@ -0,0 +1,16 @@
+from game.cards.discard_card import DiscardCard
+from game.values.locations import Locations
+import game.turn_actions.player_movement as player_movement
+
+
+class DropFromHighAltitude(DiscardCard):
+    def __init__(self):
+        super().__init__("Drop from High Altitude", 5,
+                         "+ 2[Star] and take control of Tokyo if you don't already control it")
```
cvennel: Drop from high altitude has a footnote: "(player choose to take control either City & Bay if both are occupied) "

This also means you shouldn't be moving everyone out of tokyo

gpfaia: Right now we only have 1 Tokyo, the action i see is a force field would have to happen

gpfaia:
closes #224, renames energizes correctly

backend/game/cards/discard_cards/victory_point_manipulation_cards/drop_from_high_altitude.py
cvennel: To be improved in future issue

PR #228, opened 2020-03-01
This is a small bug fix for buying cards

gpfaia:


backend/game/engine/board.py
```diff
@@ -13,7 +13,6 @@
 class BoardGame:
 
     def __init__(self):
-        self.deck_handler = DeckHandler()
```
cvennel: Why is the deckhandler going away? I think that would break consumers.py which has 7 usages of state.deck_handler

gpfaia: There is already a deckhandler on line 20, there was two in the __init__ removing one for clarity

cvennel: ha oops

PR #232, opened 2020-03-02
Store more values to db

otaiwo70:
Game, User, Dice values with Dice selected and re-rolls populate to db, DB model refactored to accommodate

DB Browser for SQLite can be used to see the data insertions


otaiwo70:
[DB_Store.pptx](https://github.com/borisv13/kot/files/4273283/DB_Store.pptx)

Recording of the DB store

backend/lobby/consumers.py
```diff
@@ -79,15 +81,21 @@ def get_or_create_user(self, username, room):
...
             game_store = i_repository_game.save_game(room)
             print("Room created with id {}".format(game_store))
         else:
             error = 'Room already exist, Unable to create Game with room: ' + room
 
+        if not i_repository_player.get_players_by_username_and_room(username, room):
+            player_store = i_repository_player.save_player(username, room)
+            print("Player created with id {}".format(player_store))
+        else:
+            error = 'Player in this room already exist, Unable to create Player in room: ' + room
```
cvennel: This isn't so much that the player can't be created, it's that the player can't be added to the database because a database entry exists. Since this database is used for analytics and not gameplay, this error message seems a little misleading. And do we want to tell the player that the background analytics aren't working? would they care?

backend/lobby/consumers.py
```diff
@@ -79,15 +81,21 @@ def get_or_create_user(self, username, room):
...
         game, created = GameState.objects.get_or_create(room_name=room)
-
         i_repository_game = IRepositoryGame()
+        i_repository_player = IRepositoryPlayer()
 
         if not i_repository_game.get_game_by_room(room):
             game_store = i_repository_game.save_game(room)
             print("Room created with id {}".format(game_store))
         else:
             error = 'Room already exist, Unable to create Game with room: ' + room
```
cvennel: See comment below, I think these error messages could be a little misleading. Approving but you may want to consider adjusting them to let us know gameplay is working but analytics are not

backend/lobby/consumers.py
otaiwo70: Great point, updated the error message to clarify

backend/lobby/consumers.py
otaiwo70: updated this error as well message to clarify

backend/lobby/consumers.py
```diff
@@ -79,14 +81,20 @@ def get_or_create_user(self, username, room):
...
+            print("Room created in analytics database with id {}".format(game_store))
+        else:
+            error = 'Room already exist in analytics database, Unable to create Game with room: ' + room
+
+        if not i_repository_player.get_players_by_username_and_room(username, room):
+            player_store = i_repository_player.save_player(username, room)
+            print("Player created in analytics database with id {}".format(player_store))
         else:
-            error = 'Room already exist, Unable to create Game with room: ' + room
+            error = 'Player in this room already exist in analytics database, Unable to create Player in room: ' + room
```
kaubster: is this really an error? If the room exists, were just getting it right?

second message, if players in room... were just utilizing past state?

this is through, but maybe not necessary?

Since the messages are passive, I'm approving.

Issue #234, opened 2020-03-03
Post_roll_actions in Board.py


cvennel:
The post Roll actions call the special effect of a card but not through an instance of the card so "self" is not defined. Therefore the game thinks the other players are not defined since every arg is shifting left one.

Applies to both SolarPowered and EnergyHoarder

  File ".....\backend\game\engine\board.py", line 61, in post_roll_actions
    active_player, self.players.get_all_alive_players_minus_current_player())
  special_effect() missing 1 required positional argument: 'other_players'

PR #235, opened 2020-03-03
Recover failing build

cvennel:
recovered missing render() portion from Gameboard.js

Commented out problematic card code from board.py
see issue #234 for details on board.py

### DIFF
--- a/backend/game/engine/board.py
+++ b/backend/game/engine/board.py
@@ -1,4 +1,6 @@
 import game.values.constants as constants
+from game.cards.keep_cards.energy_manipulation_cards.energy_hoarder import EnergyHoarder
+from game.cards.keep_cards.energy_manipulation_cards.solar_powered import SolarPowered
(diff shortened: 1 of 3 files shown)


backend/game/engine/board.py
```diff
@@ -57,12 +54,12 @@ def re_roll(self, indexes_to_re_roll):
 
     def post_roll_actions(self, active_player):
         # TODO, refactor to other method
-        if active_player.has_instance_of_card(EnergyHoarder()):
-            EnergyHoarder.special_effect(
-                active_player, self.players.get_all_alive_players_minus_current_player())
```
gpfaia: I think if you either set the card as an instance and call special effect, it should work as intended

`EnergyHoarder().special_effect(active_player, self.players.get_all_alive_players_minus_current_player())`

cvennel: Initialized as suggested. This PR now fixes #234

PR #238, opened 2020-03-06
Drop from high altitude support

### LINKED ISSUE #227: Need a "force" move out of Toyko
The card 'drop from high altitude' #95 has functionality where if the player is not in Tokyo already, it can remove the the player and they take over.  Basically a force field.  It gets even more complicated when they get the choice to force yield Tokyo city or bay if both are present.

cvennel:
This adds options to choose who to force to yield in the event multiple people are in Tokyo

closes #227


cvennel:
2PlayerDemo

cvennel:
ezgif com-video-to-gif (4)

PR #240, opened 2020-03-06
Additional db store

otaiwo70:
storing card sweep, card buy and card use to DB


backend/game/deck/deck_handler.py
```diff
@@ -94,6 +95,7 @@ def buy_card_from_store(self, index, purchasing_player: Player, other_players):
     # optional second arg allows discard to handle removing card from origin as well
     def discard(self, card: Card, card_from_location: List[Card] = None):
         self.__discard_pile.append(card)
+        self.cards_swept.append(card)
```
cvennel: I don't think this is where you want to get the card swept info. There's a method specifically to sweep the card store. Sweeping gets rid of all cards. I think this discard method  is called any time a discard card is purchased. This would give you the wrong data. You probably want to handle the DB stuff at the Consumers.py level anyways, then whenever the card sweep request is called, just grab the current state of card store using `cards_to_sweep = state.deckhandler.card_store` or whatever it is right before actually sweeping

otaiwo70: Thanks for those details they were very helpful, took a different approach thou in that i capture the top 3 card in deck_handler and then just pass over to consumers.py, and the solution works as needed

backend/game/deck/deck_handler.py
```diff
@@ -29,6 +29,7 @@ def __init__(self):
         self.__draw_pile.get_new_deck()
         self.__draw_pile.shuffle()
         self.__fill_card_store()
+        self.cards_swept = []
```
cvennel: not needed

backend/game/deck/deck_handler.py
```diff
@@ -106,3 +107,9 @@ def sweep_store(self, player_invoking_sweep: Player):
             for _ in range(3):
                 self.discard(self.store.pop())
             self.__fill_card_store()
+
+    def get_top_three_cards(self):
```
cvennel: not needed

backend/lobby/consumers.py
```diff
@@ -354,8 +358,26 @@ def buy_card_request_handler(self, data):
     def card_store_sweep_request_handler(self, data):
         username, room, game, state = reconstruct_game(data)
         if state.players.current_player.username == username:
+            cards_swept = state.deck_handler.get_top_three_cards()
```
cvennel: The get_top_three_cards method could be removed.
`cards_swept = state.deck_handler.cardstore.copy()`

otaiwo70: Awesome and appreciate the feedback, the code had a slight issue but it worked as 

cards_swept = state.deck_handler.store.copy()

will need your assistance as an additional approval to merge

PR #244, opened 2020-03-09
Btn gray roll

### LINKED ISSUE #233: All buttons grayed out if AcceptDice before roll
Clicking Roll before accept locks all dice

### LINKED ISSUE #239: Turn indicator doesn't change until after accepting dice

kaubster:
Turn indicator doesn't change until after accepting dice #239
All buttons grayed out if AcceptDice before roll #233

closes #233
closes #239


setup_build_env.bat
```diff
@@ -0,0 +1,14 @@
+python -m venv env
```
gpfaia: Should this file be in here?

PR #254, opened 2020-03-15
Add Rooting for Underdog resolves #134

### LINKED ISSUE #134: Rooting for the Underdog
Cost: 3
Type: Keep
Effect: At the end of a turn if you have the fewest [Star] gain 1 [Star].
Footnote: (no [star] if tied fewest; at the end of your turn)

gpfaia:


backend/game/engine/board.py
```diff
@@ -62,6 +63,10 @@ def post_roll_actions(self, active_player):
         if active_player.has_instance_of_card(SolarPowered()):
             SolarPowered().special_effect(
                 active_player, self.players.get_all_alive_players_minus_current_player())
+        if active_player.has_instance_of_card(RootingForTheUnderdog()):
```
cvennel: Seems like this would be better in the get_next_player_turn since cards could cause the player to no longer have the fewest stars between post_roll and next_turn

gpfaia: Originally was going to move it but all of these are post_roll/end of turn cards so even though the function may not be named properly this is the end of turn and the next function call is always to get the next player's turn

cvennel: Oh ok

